PR #3, opened 2019-08-30
Initial Dart backend implementation

mit-mit:


internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+	return pkgs, true
+}
+
+// DartPubBackend is a UPM backend for Dart that uses Pub.dev.
+var DartPubBackend = api.LanguageBackend{
+	Name:             "dart-pub",
+	Specfile:         "pubspec.yaml",
+	Lockfile:         "pubspec.lock",
+	FilenamePatterns: []string{"*.dart"},
+	Quirks:           api.QuirksNone,
```
raxod502: If the same command is used for both lock and install, then maybe QuirksLockAlsoInstalls would be appropriate?

internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+
+	writeSpecFile(specs)
+}
+
+// dartGuess stub.
+func dartGuess() (map[api.PkgName]bool, bool) {
+	pkgs := map[api.PkgName]bool{}
+
+	return pkgs, true
+}
```
raxod502: It's fine to panic with NotImplemented for this, I think. The fact that it isn't implemented can be documented in the README.

internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+		HomepageURL:   "",
+		SourceCodeURL: "",
+		BugTrackerURL: "",
+		Author: util.AuthorInfo{
+			Name:  "",
+			Email: "",
+			URL:   "",
+		}.String(),
+		License: ""}
+}
```
raxod502: If the API doesn't return this info, Name at least should be filled out so as to satisfy the UPM API contract.

internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+		util.Die("Pub.dev: %s", err)
+	}
+
+	results := make([]api.PkgInfo, len(pubDevResults.Packages))
+	for i := range pubDevResults.Packages {
+		p := pubDevResults.Packages[i]
+		results[i] = api.PkgInfo{
+			Name: p.Name,
+		}
+	}
```
raxod502: Is this really all the info the API returns?

internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+	return pkgs
+}
+
+// ParseDartSpec parses a Dart version handling special
+// cases like Flutter SDK deps where the spec is a map.
+func dartParseSpec(spec interface{}) string {
+	version, ok := spec.(string)
+	// Handle cases like Flutter SDK deps where the spec is a map.
+	if !ok {
+		version = "Unknown"
```
raxod502: Maybe empty string would be better?

mit-mit:
cc @jonasfj to review API usage

internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+type pubDevSearchResults struct {
+	Packages []struct {
+		Name string `json:"package"`
+	} `json:"packages"`
+	NextURL string `json:"next"`
+}
+
+// dartSearch implements Search for Pub.dev.
+func dartSearch(query string) []api.PkgInfo {
+	endpoint := "https://pub.dev/api/search"
```
jonasfj: Make a global const `pubBaseUrl`

```golang
import "os"

// Base URL for pub repository.
//
// Still using pub.dartlang.org (instead of pub.dev), because this is used by the pub client, and
// by implication necessarily stable.
const defaultPubBaseUrl = "https://pub.dartlang.org"

String getPubBaseUrl() {
  baseUrl := os.Getenv("PUB_HOSTED_URL")
  if baseUrl != "" {
    return baseUrl
  }
  return defaultPubBaseUrl
}

func dartSearch (query string) []api.PkgInfo {
    endpoint := fmt.Sprintf("%s/api/search", pubBaseUrl);

   // ...
}
```

internal/backends/dart/dart.go
mit-mit: I'm afraid so: https://pub.dev/api/search?q=protobuf

internal/backends/dart/dart.go
mit-mit: This is now implemented!

internal/backends/dart/dart.go
```diff
@@ -0,0 +1,250 @@
...
+	queryParams := "?q=" + url.QueryEscape(query)
+
+	resp, err := http.Get(endpoint + queryParams)
```
jonasfj: Please add a custom user-agent:

```golang
req, err := NewRequest("GET", endpoint + queryParams, nil)
if err != nil {
  util.Die("Pub.dev: %s", err)
}
// Add custom User-Agent which helps find the source of requests.
req.Header.Add("User-Agent", "upm (+https://github.com/replit/upm)")
req.Header.Add("Accept", "application/json")
res, err := http.DefaultClient.Do(req)
if err != nil {
  util.Die("Pub.dev: %s", err)
}

//....
```

internal/backends/dart/dart.go
mit-mit: Done

internal/backends/dart/dart.go
mit-mit: Done

internal/backends/dart/dart.go
mit-mit: Done

PR #4, opened 2019-09-04
better python package guessing

turbio:
pipreqs doesn't quite get the job done reliably for what we're trying to do.

Instead we've gone and generated our own mapping of every pypi package (that poetry can install) to the modules it makes available. This module list intentionally includes the modules of transitive dependencies.


raxod502:
Why does it include transitive dependencies?

turbio:
@raxod502 they're used over here: https://github.com/replit/upm/pull/6
it doesn't make a whole lot of sense for the guessing step but it seems to give better results than pipreqs. If I have time separate transitive/direct modules them in the future.

turbio:
follow up issue https://github.com/replit/upm/issues/7

PR #10, opened 2019-10-31
Add installation script

kognise:
I added a method for more easily installing UPM on Debian-based Linux systems. The script will automatically pull and install the latest version of UPM from GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ tap](https://docs.brew.sh/Taps).
 .deb packages are available on the [releases
 page](https://github.com/replit/upm/releases).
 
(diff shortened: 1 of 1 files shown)


raxod502:
This seems like a neat idea to me, but the file should definitely be hosted in this repository instead of in an unrelated Gist. Would it be possible to support some of the other installation methods, like at least RPM and Homebrew, and choose between them based on the operating system?

blast-hardcheese:
Agreed! Very neat, out of scope for the time being though. I think we'd need to implement this ourselves, though I'm converting this PR to an issue to track. Thanks for the suggestion!

PR #13, opened 2019-11-01
[PKGBUILD] Ensure it follows guidelines

Foxboron:
Binary packages should have the -bin suffix if the source is available.
Because it's a precompiled binary, this won't work on any architecture,
thus we change it to `x86_64`. License is also unique, so added to the
license directories.

I see you have already uploaded the package to AUR [with a bot](https://aur.archlinux.org/packages/upm/). Needs to have `upm-bin`  created and then we can do a merge request to rename the package. 

Please ask if there is any questions regarding packaging and/or AUR.

### DIFF
--- a/packaging/aur/.SRCINFO
+++ b/packaging/aur/.SRCINFO
@@ -1,11 +1,12 @@
-pkgbase = upm
+pkgbase = upm-bin
 	pkgdesc = Universal package manager: Python, Node.js, Ruby, Emacs Lisp.
 	pkgver = 1.0
 	pkgrel = 1
-	arch = any
+	url = https://repl.it/site/blog/upm
+	arch = x86_64
 	license = MIT
 	source = https://github.com/replit/upm/releases/download/v1.0/upm_1.0_linux_amd64.tar.gz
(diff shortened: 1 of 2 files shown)


raxod502:
This is super helpful, thanks a lot! This was the first time I created a package for AUR.

It's worth noting that we could probably also create a source package pretty easily, since it just needs to depend on Go and then build a single binary.

Foxboron:
There are guidelines for golang packaging which should help you a lot on the way.

https://wiki.archlinux.org/index.php/Go_package_guidelines

dan-stowell:
Good afternoon @Foxboron ! You've probably noticed by now that we've been slow to merge community PRs on polygott. We've decided to focus on giving Repl.it [a stable foundation](https://blog.repl.it/quality) instead of adding features and updating languages. Once we and our community feel comfortable with Repl.it's overall stability, we'll give this change another look. Thank you for contributing!

Foxboron:
Okay, but that means I'll be submitting a deletion request on the AUR package until it follows the packaging guidelines.

raxod502:
@dan-stowell I would definitely suggest merging this rather than waiting!

PR #14, opened 2019-11-01
Merge the conditions to remove duplicate code

DanielRuf:
This PR merges the if / else if into one as the body of both branches of the condition is the same.

### DIFF
--- a/resources/nodejs/babel-parser.js
+++ b/resources/nodejs/babel-parser.js
@@ -4168,9 +4168,7 @@ var typescript = (superClass => class extends superClass {
     this.expect(types.parenL);
     signature.parameters = this.tsParseBindingListForSignature();
 
-    if (returnTokenRequired) {
-      signature.typeAnnotation = this.tsParseTypeOrTypePredicateAnnotation(returnToken);
-    } else if (this.match(returnToken)) {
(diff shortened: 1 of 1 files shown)


raxod502:
This is vendored code so we probably shouldn't be touching it. It's taken directly from the distribution copy of Babel.

DanielRuf:
So why is it in the sources / tracked then?

raxod502:
Because the build system would be a lot more complicated and fragile otherwise.

DanielRuf:
Hm, so no lockfiles? And we should not touch these files? So we can put them on the .gitignore file then to prevent that others touch them too?

DanielRuf:
Hm, so no lockfiles? And we should not touch these files? So we can put them into the .gitignore file then to prevent that others touch them too?

raxod502:
Your logic would be perfect in a Node.js project. There, it would make perfect sense to use the package manager to pull in the correct version of Babel. However, this isn't a Node.js project. It's a Go project, which needs to vendor code from Node.js, from Python, from Ruby, etc. It's simply not practical or desirable to say that in order to build this project, you need to install every package manager on your machine.

DanielRuf:
No I meant that the vendor files can be added to the .gitignore file after adding them to prevent that others do the same mistake like I did.

PR #11, opened 2019-10-31
Update Scoop installation in README.md

r15ch13:
upm has been added to the main bucket with https://github.com/ScoopInstaller/Main/commit/029935dd01dada7562f423bcc54e54fc9490bc50

PR #21, opened 2019-11-11
Mention upm available in the Scoop main bucket

JanPokorny:
UPM is now available in the Scoop main bucket (added by default): https://github.com/ScoopInstaller/Main/blob/master/bucket/upm.json


JanPokorny:
Ooops, didn't check for dupes (#11), my bad

PR #25, opened 2020-03-05
add java-maven backend

dan-stowell:
This change adds a `java-maven` backend that can add, remove, and search packages from the Maven central repository.

TODOs:
- Have `upm -l java info <packagename>` return the package description if it exists.
- search will only return the latest version for a package. A package such as `com.google.guava:guava` has a `28.2-jre` and `28.2-android` latest version - it'd be nice to return multiple versions so that folks can add the correct one for their project.

### DIFF
--- a/internal/api/types.go
+++ b/internal/api/types.go
@@ -315,30 +315,39 @@ type LanguageBackend struct {
 // a builder function which can perform this normalization and
 // validation.
 func (b *LanguageBackend) Setup() {
-	if b.Name == "" ||
-		b.Specfile == "" ||
-		b.Lockfile == "" ||
-		len(b.FilenamePatterns) == 0 ||
-		b.GetPackageDir == nil ||
-		b.Search == nil ||
-		b.Info == nil ||
-		b.Add == nil ||
-		b.Remove == nil ||
+	condition2flag := map[string]bool{
+		"b.Name == \"\"": b.Name == "",
+		"b.Specfile == \"\"": b.Specfile == "",
+		"b.Lockfile == \"\"": b.Lockfile == "",
+		"len(b.FilenamePatterns) == 0": len(b.FilenamePatterns) == 0,
+		"b.GetPackageDir == nil": b.GetPackageDir == nil,
+		"b.Search == nil": b.Search == nil,
+		"b.Info == nil": b.Info == nil,
+		"b.Add == nil": b.Add == nil,
+		"b.Remove == nil": b.Remove == nil,
 		// The lock method should be unimplemented if
 		// and only if builds are not reproducible.
-		((b.Lock == nil) != b.QuirksIsNotReproducible()) ||
-		b.Install == nil ||
-		b.ListSpecfile == nil ||
-		b.ListLockfile == nil ||
+		"((b.Lock == nil) != b.QuirksIsNotReproducible())": ((b.Lock == nil) != b.QuirksIsNotReproducible()),
+		"b.Install == nil": b.Install == nil,
+		"b.ListSpecfile == nil": b.ListSpecfile == nil,
+		"b.ListLockfile == nil": b.ListLockfile == nil,
 		// If the backend isn't reproducible, then lock is
 		// unimplemented. So how could it also do
 		// installation?
-		b.QuirksDoesLockAlsoInstall() &&
-			b.QuirksIsNotReproducible() ||
+		"b.QuirksDoesLockAlsoInstall() && b.QuirksIsNotReproducible()": b.QuirksDoesLockAlsoInstall() && b.QuirksIsNotReproducible(),
 		// If you install, then you have to lock.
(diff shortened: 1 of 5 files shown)


internal/api/types.go
```diff
@@ -315,30 +315,39 @@ type LanguageBackend struct {
...
-	if b.Name == "" ||
-		b.Specfile == "" ||
-		b.Lockfile == "" ||
-		len(b.FilenamePatterns) == 0 ||
-		b.GetPackageDir == nil ||
-		b.Search == nil ||
-		b.Info == nil ||
-		b.Add == nil ||
-		b.Remove == nil ||
+	condition2flag := map[string]bool{
```
turbio: could you make the reason strings a bit more friendly. At least remove the `b.`.

internal/backends/java/java.go
```diff
@@ -0,0 +1,201 @@
...
+	ModelVersion string       `xml:"modelVersion"`
+	GroupId      string       `xml:"groupId"`
+	ArtifactId   string       `xml:"artifactId"`
+	Version      string       `xml:"version"`
+	Dependencies []Dependency `xml:"dependencies>dependency"`
+}
+
+var emptyProject = Project{
+	ModelVersion: "4.0.0",
+	GroupId:      "co.repl",
```
turbio: hmm, unsure what the default value should be but people using upm outside of replit might not want co.repl

internal/backends/java/java.go
```diff
@@ -0,0 +1,201 @@
...
+			submatches := pkgNameRegexp.FindStringSubmatch(string(pkgName))
+			if nil == submatches {
+				util.Die("package name %s does not match groupid:artifactid pattern")
+			} else {
+				groupId := submatches[1]
+				artifactId := submatches[2]
+				if _, ok := existingDependencies[pkgName]; ok {
+					// this package is already in the lock file
+				} else {
+					var versionString string
```
turbio: things get really nested in here, try to avoid deep indentation. We **generally** try to stay <= 80 columns with tabs being 4 spaces.

internal/backends/java/java.go
```diff
@@ -0,0 +1,201 @@
...
+	Lockfile:         pomdotxml,
+	FilenamePatterns: javaPatterns,
+	Quirks:           api.QuirksAddRemoveAlsoLocks,
+	GetPackageDir: func() string {
+		return "target/dependency"
+	},
+	Search: func(query string) []api.PkgInfo {
+		searchDocs, err := Search(query)
+		if err != nil {
+			util.Log("error searching maven %s", err)
```
turbio: what's the reason for logging but not exiting with an error?

internal/backends/java/java.go
```diff
@@ -0,0 +1,201 @@
...
+				artifactId := submatches[2]
+				if _, ok := existingDependencies[pkgName]; ok {
+					// this package is already in the lock file
+				} else {
+					var versionString string
+					if pkgSpec == "" {
+						searchDocs, err := Search(fmt.Sprintf("g:%s AND a:%s", groupId, artifactId))
+						if err != nil {
+							util.Log("error searching maven for latest version of %s:%s: %s", groupId, artifactId, err)
+							return
```
turbio: should this be an error?

internal/backends/java/java.go
```diff
@@ -0,0 +1,201 @@
...
+						}
+						if len(searchDocs) == 0 {
+							util.Log("did not find a package %s:%s", groupId, artifactId)
+							return
```
turbio: and this???

internal/backends/java/java.go
dan-stowell: going with `mygroupid` and `myartifactid`

internal/api/types.go
dan-stowell: yes! thank you for the reminder.

internal/backends/java/java.go
dan-stowell: no reason! changing many of these to `util.Die()`

internal/backends/java/java.go
dan-stowell: yes!

internal/backends/java/java.go
dan-stowell: I pulled out functions longer than a line and broke up some looong lines.

internal/backends/java/java.go
dan-stowell: changed this to `Die` as well

dan-stowell:
> pls run gofmt locally

Ah, yes, some of these changes I edited right in repl.it, some in VS Code. Everything should be formatted now.

turbio:
nope, that's exactly the right flow :+1:

internal/api/types.go
turbio: now these are some top notch errors :ok_hand:

PR #27, opened 2020-03-10
root flag to expand ignored paths

turbio:
why
===

Some frameworks/templates could be made a bit faster by explicitly ignoring
known paths from upm's guess. These paths aren't general enough to apply to all
languages or any specific language in all cases. To give the end user (who
hopefully knows best) more control we'll add an ignored paths flag.


turbio:
@masad-frost yup, should i change the help text to be more clean?

PR #29, opened 2020-03-24
Tag successful builds of master

dan-stowell:
So that polygott can explicitly depend on different versions of upm, let's tag each successful `build_test_and_deploy` of master!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   build_and_test:
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,28 +1,51 @@
...
+      - run: apk add --no-cache --no-progress make git openssh
+      - run:
+          name: Keyscan Github (HACK) #https://discuss.circleci.com/t/known-hosts-in-circle-2-0/18544
+          command: ssh-keyscan -H github.com >> ~/.ssh/known_hosts
       - checkout
       - setup_remote_docker
       - run: >-
           echo "$DOCKER_PASSWORD" |
             docker login --username "$DOCKER_USERNAME" --password-stdin
       - run: apk add --no-cache --no-progress make
```
kochman: I think this line is redundant now

dan-stowell: Ah yes, thank you!

PR #30, opened 2020-03-26
Add R support

ALANVF:
An attempt to add package manager support for the R programming language. There's a good chance that it won't work on here but it worked when testing it on repl.it.

## How the package manager works
So R doesn't actually have a real package manager, but instead provides functions that you can use instead. I decided to use/make Go bindings for R to make this work in Go. I mostly use the functions `install.packages` and `find.package`. As for the lock/spec files, they're just JSON files that keep track of the packages that are installed (R doesn't require them, but they were the easiest option). I also tried to have it keep track of package versions, although I can't figure out how to make that work because there is little-to-no documentation on their packages website for choosing package versions (I could try to look into it further, but this works for now).


internal/backends/rlang/rlang.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/replit/upm/internal/util"
+)
+
+func getImports(imports string) []string {
+	return regexp.MustCompile(`[a-zA-Z_]\w*`).FindAllString(imports, -1)
+}
+
+func createRPkgDir() {
+	if _, err := os.Stat("~/R"); os.IsNotExist(err) {
+		if err := os.MkdirAll("~/R/x86_64-pc-linux-gnu-library/3.4", os.ModeDir); err != nil {
```
dan-stowell: We often put packages in the working directory of the running process. Intermediate stuff we can go elsewhere (for example, Maven uses `~/.m2/repository`). It really depends on the convention for the package manager!

ALANVF: Ok I can change that. I had done that because that is the directory that R defaults to if it can't install anywhere else

internal/backends/rlang/specfile.go
```diff
@@ -0,0 +1,166 @@
...
+		if installed.Name == pkg.Name {
+			return true
+		}
+	}
+
+	return false
+}
+
+func RAdd(pkg RPackage) {
+	if file, err := os.Open("~/Rconfig.json"); err == nil {
```
masad-frost: Same comment as @dan-stowell had on using the working directory. If the config is on the homedir then all the projects will share the same config

internal/backends/rlang/specfile.go
```diff
@@ -0,0 +1,166 @@
...
+			file.Close()
+			panic(err)
+		}
+	} else {
+		panic(err)
+	}
+}
+
+func RGetLockFile() RConfig {
+	if file, err := os.Open("~/Rconfig.json.lock"); err == nil {
```
masad-frost: If this is an actual json file maybe a better extension might be `.lock.json`, then you end up with a json mime type

internal/backends/rlang/specfile.go
masad-frost: Make sure you do this for all other files, I see you addressed it only for the `~/R` directory

ALANVF: Should be fixed now

internal/backends/rlang/specfile.go
ALANVF: Done

ALANVF:
@masad-frost I have updated my description. Let me know if you need further clarification

internal/backends/rlang/rlang.go
```diff
@@ -0,0 +1,138 @@
+package rlang
+
+import (
+	"fmt"
+	"regexp"
+	"github.com/ALANVF/rgo"
```
dan-stowell: When running `make full` locally on your branch, I encounter this error: 

```
cd cmd/upm && go build -ldflags "-X 'github.com/replit/upm/internal/cli.version=${VERSION:-development version}'"
# github.com/ALANVF/rgo
/root/go/pkg/mod/github.com/!a!l!a!n!v!f/rgo@v0.0.0-20200324171131-7f868d8e11f2/rgo.go:11:10: fatal error: R.h: No such file or directory
 #include <R.h>
          ^~~~~
compilation terminated.
make: *** [Makefile:12: cmd/upm/upm] Error 2
```

Building `rgo` depends on having installed the R headers. Currently you do that in `docker-install-languages.bash` (understandable!). Can you try installing `r-base-dev` in `docker-install-build.bash` and see if you're able to run `make full`?

ALANVF: Yes I can try that. I had developed `rgo` on repl.it since I couldn't figure out how to use the R headers locally on windows, so I wasn't sure what dependencies were needed to make it work here

ALANVF: Fixed now

PR #32, opened 2020-04-18
Implemented Info for java-maven

navteniev:
Added Info function to maven central search and introduced a few tests.

### DIFF
--- a/internal/backends/java/msch.go
+++ b/internal/backends/java/msch.go
@@ -25,8 +25,11 @@ package java
 
 import (
 	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/backends/java/msch.go
```diff
@@ -46,18 +50,45 @@ type SearchResult struct {
...
 }
+
+func Search(keyword string) ([]SearchDoc, error) {
+	searchURL := mavenURL + url.QueryEscape(keyword)
+
+	return mavenSearch(searchURL)
+}
+
+func Info(name string) (SearchDoc, error) {
+	parts := strings.Split(string(name), ":")
```
dan-stowell, line 82: Can you make this resilient to badly-formatted names? For example, running `upm -l java info blade-mvc` produces a stack trace because I've left out a group name.

navteniev: Yeah, let me see what I can do.

navteniev: Changed the implementation to treat input without a group as artifact only search.

navteniev:
@dan-stowell I've made the changes to handle missing group and the result:
`
upm info blade-mvc
Name:      com.bladejava:blade-mvc
Version:   2.0.15.RELEASE
`

Give it a try.

PR #33, opened 2020-04-23
Switch to using esbuild's parser, removing the 1s babel overhead

amasad:
Switched to using ESBuild's parser for nodejs instead of babel which has almost a one second overhead between booting up nodejs and reading and parsing the files in babel. This reduces the parsing overhead to something negligible. 

Generally 1s upm run vs .2s

Screen Shot 2020-04-22 at 11 27 46 PM


turbio:
looks like master if failing :/

amasad:
@turbio you or @dan-stowell should take over. I wrote it in repl.it and I don't think the tests run there

turbio:
I can review but I don't have time to take this one right now

dan-stowell:
@amasad @turbio this is on my list for today - have to get polygott to fetch and build tags of UPM first.

amasad:
Btw, one of the reasons upm is adding a fixed cost to React.js projects is because it's failing to parse JSX, which this pull request fixes too

amasad:
@dan-stowell feel free to rewrite in a more Go idiomatic way haha. I should have maybe made it into a map.

The other thing that I think I could've done is use filepath walker instead of implementing my own file/dir walker

masad-frost:
Looks like you read my pending review from yesterday 

image

PR #36, opened 2020-04-30
Ignore loaders and absolute paths

amasad:
Webpack allows users to add `!` to load arbitrary files and this should be ignored by upm. 
More generally I ported the code from here https://github.com/replit/upm/blob/c9627d248a515cbe594950227afd5f1b2600b867/resources/nodejs/bare-imports.js#L160-L177


amasad:
Thanks!

Issue #34, opened 2020-04-24
Add support for Clojure


vhawk19:
From what i understand prybar-clojure uses clojure-cli tools and deps.edn, would upm support clojure librarie either using deps.edn or lein(using project.clj). I see that as my primary limiting factor while using clojure on repl.it unlike python/dart where I can import libraries with ease.

fctorial:
deps.edn requires an alias to be defined in order to run a clojure program. Should a deps project initialized with upm add this alias? Should upm create the main source file for that alias as well? i.e. Generate a `void main() ` for clojure.

Billcab:
Having trouble creating alias as I am trying to digitize object of file conflicts in programming languages

ryantm:
We are open to new backend pull requests for popular languages if they come with tests, and we do not need to leave these issues open to track.

PR #39, opened 2020-05-29
Deps

fctorial:
Adds support for clojure via tools.deps. Addresses #34

### DIFF
--- a/internal/backends/clojure/deps.go
+++ b/internal/backends/clojure/deps.go
@@ -2,6 +2,7 @@ package clojure
 
 import (
 	"encoding/json"
+	"encoding/xml"
 	"errors"
 	"github.com/replit/upm/internal/api"
 	"github.com/replit/upm/internal/util"
@@ -36,6 +37,8 @@ func (self *JavaPackage) FromString(s string) error {
 	return nil
 }
 
+// Info obtained from maven/clojar are coerced into this
+// and this struct is coerced into api.PkgInfo
 type MPkgInfo struct {
 	Id            JavaPackage
 	LatestVersion string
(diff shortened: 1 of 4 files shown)


internal/backends/clojure/deps.go
```diff
@@ -0,0 +1,504 @@
...
+		if err != nil {
+			panic(err)
+		}
+		fres = append(fres, GenPkgInfo(pinfo))
+	}
+
+	return fres, nil
+}
+
+func findPackagesMaven(name string) ([]api.PkgInfo, error) {
```
dan-stowell, line 249: Neat! I adapted some code to search Maven for the `java-maven` backend: https://github.com/replit/upm/blob/master/internal/backends/java/msch.go - I like what you've written here though and in a follow-on change I may point the `java-maven` backend at these structs & functions!

internal/backends/clojure/deps.go
```diff
@@ -0,0 +1,504 @@
...
+func Search(name string) []api.PkgInfo {
+	cpkg, err := findPackagesClojars(name)
+	if err != nil {
+		util.Log("Error while querying clojars: " + err.Error())
+	}
+	mpkg, err := findPackagesMaven(name)
+	if err != nil {
+		util.Log("Error while querying maven: " + err.Error())
+	}
+	return append(mpkg, cpkg...)
```
dan-stowell, line 286: Is it possible for a single package to come back from clojars and from maven? If so, we'll want to remove duplicates.

internal/backends/clojure/deps.go
```diff
@@ -0,0 +1,504 @@
...
+	for _, entry := range traceEdn.Log {
+		res[api.PkgName(string(entry.Lib))] = api.PkgVersion(entry.UsedVersion.Version)
+	}
+
+	return res
+}
+
+func install() {
+	util.Log("installing, this will download the dependencies")
+	util.RunCmd([]string {"clj", "--eval", "(System/exit 0)"})
```
dan-stowell, line 503: I'm not terribly familiar with clojure - will this command automatically run `src/main.clj`? What if a project has `./main.clj` and `src/main.clj`?

fctorial: `clj --eval "(System/exit 0)"` will download the dependencies, generate the classpath and launch an repl with that classpath which will exit immediately. It only needs `deps.edn` to work.

internal/backends/clojure/deps.go
fctorial: Yes, I'll get it done.

dan-stowell:
Thank you for removing duplicates! I have one last request: can you install clojure in the docker-install-languages.bash script? I went to test this out by running `make upm; make dev`, and `clj` was not on the path in the container!

Here's how we install clojure in polygott: https://github.com/replit/polygott/blob/master/languages/clojure.toml#L13

(Ideally we would have tests that pointed out stuff like this.)

fctorial:
I've changed the command name to `clojure`. For some reason `clj` isn't present in ubuntu clojure package, although it is there in upstream package:

```
[$] tar tf ~/Downloads/clojure-tools-1.10.1.492.tar.gz                                                                                                                            
clojure-tools/
clojure-tools/clojure-tools-1.10.1.492.jar
clojure-tools/install.sh
clojure-tools/clj.1
clojure-tools/example-deps.edn
clojure-tools/clojure
clojure-tools/clojure.1
clojure-tools/clj
clojure-tools/deps.edn
```

So now upm works in docker container.

dan-stowell:
Good afternoon @fctorial ! You've probably noticed by now that we've been slow to merge community PRs on polygott. We've decided to focus on giving Repl.it [a stable foundation](https://blog.repl.it/quality) instead of adding features and updating languages. Once we and our community feel comfortable with Repl.it's overall stability, we'll give this change another look. Thank you for contributing!

PR #42, opened 2020-06-04
create virtualenv under .upm for python

dan-stowell:
The problem I'm trying to solve: in a repl, I'd like to create a virtualenv in the working directory only when someone adds a python package.

upm knows when it's adding a python package, so I figured it could create the virtualenv at `.upm/virtualenvs/python(2|3)/`! If you don't want upm to create a virtualenv, you can point to an existing virtualenv with the `VIRTUAL_ENV` env var.

If creating virtualenvs falls outside upm's remit, I'm happy to move this logic elsewhere!

This change also explicitly installs poetry 1.0.5 (we've been implicitly pulling that version of poetry for some time).

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -3,6 +3,7 @@ package python
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


turbio:
I'm thinking this should be outside upm's jurisdiction but i'm not sure. Setting up a venv is tricky as the rest of the environment has to agree and use the venv, not just upm. We'd somehow bubble up this information and at that point I'm not sure it makes sense to put in upm.

On a replit specific note: pretty sure we don't want to be initializing a fresh venv into everyone's project dir. It's about 8mb and locks the environment in ways that will be tricky for us to smoothly migrate/upgrade. I think we'd ideally just include a super stripped site-packages.

dan-stowell:
Thanks for the thoughts, as always, @turbio! I'll close this and we can figure out the right path somewhere other than a PR.

aemonge:
Considering that there's no other solution out there, I came across UPM looking for one ring to manage all virtual envs (rbenv, conda, nodenv). 

I wanted to know if this is something to look forward in UPM, or if you have any other multiple virtual environment manager alternatives.

Many Thanks, :pray:  in advance :)

PR #43, opened 2020-07-07
Fix R path issue

ALANVF:
Just a small patch.

### DIFF
--- a/internal/backends/rlang/rlang.go
+++ b/internal/backends/rlang/rlang.go
@@ -24,10 +24,8 @@ func createRPkgDir() {
 
 func updateLibPaths() {
 	util.RunCmd([]string{
-		"R",
-		"--no-echo",
-		"-e",
-		`".libPaths('./R/x86_64-pc-linux-gnu-library/3.4')"`,
+		"export",
+		`R_LIBS_USER=./R/x86_64-pc-linux-gnu-library/3.4`,
 	})
 }
 
--- a/internal/backends/rlang/rlang.go
+++ b/internal/backends/rlang/rlang.go
@@ -3,6 +3,7 @@ package rlang
 import (
 	"os"
 	"regexp"
+	"path"
 
 	"github.com/replit/upm/internal/api"
 	"github.com/replit/upm/internal/util"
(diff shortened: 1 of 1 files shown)


internal/backends/rlang/rlang.go
```diff
@@ -24,10 +24,8 @@ func createRPkgDir() {
 
 func updateLibPaths() {
 	util.RunCmd([]string{
-		"R",
-		"--no-echo",
-		"-e",
-		`".libPaths('./R/x86_64-pc-linux-gnu-library/3.4')"`,
+		"export",
```
dan-stowell, line 27: I hate to say it, but the `R_LIBS_USER` environment variable will not live past this `RunCmd` call.  If different calls to `R` need `R_LIBS_USER`, you will have to pass it through in the environment.

ALANVF: Should be fixed now

internal/backends/rlang/rlang.go
```diff
@@ -13,33 +14,35 @@ func getImports(imports string) []string {
 }
 
 func createRPkgDir() {
-	if _, err := os.Stat("./R"); os.IsNotExist(err) {
-		if err = os.MkdirAll("./R/x86_64-pc-linux-gnu-library/3.4", os.ModeDir); err != nil {
+	home := os.Getenv("HOME")
+
+	if _, err := os.Stat(path.Join(home, "R")); os.IsNotExist(err) {
+		rPath := path.Join(home, "R", "x86_64-pc-linux-gnu-library", "3.4")
```
dan-stowell, line 20: Some package managers in upm install into a directory under $HOME, some install elsewhere, and some install into the current working directory. I'm not sure what the most friendly choice for R is - under $HOME is fine and dandy. But you don't _have to_ install under $HOME if you don't want to.

ALANVF: Yeah, I had just done that because this is where R installs packages if it doesn't have perms to install to the normal directory (located somewhere in /usr/lib I think).

internal/backends/rlang/rlang.go
```diff
@@ -13,33 +14,35 @@ func getImports(imports string) []string {
...
+		if rLibsUser := os.Getenv("R_LIBS_USER"); rLibsUser != "" {
+			rPath = rLibsUser
+		}
+
+		if err = os.MkdirAll(rPath, os.ModeDir + os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
ALANVF: Done

internal/backends/rlang/rlang.go
```diff
@@ -83,39 +86,45 @@ var RlangBackend = api.LanguageBackend{
...
 				Name:    string(name),
 				Version: string(info),
 			}
 			RAdd(pkg)
+			
+			util.RunCmd([]string{
+				"R",
+				"--no-echo",
+				"-e",
+				"if(length(find.package('" + pkg.Name + "', quiet=T)) == 0) install.packages('" + pkg.Name + "')",
```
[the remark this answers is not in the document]
ALANVF: Done

internal/backends/rlang/rlang.go
```diff
@@ -12,38 +14,72 @@ func getImports(imports string) []string {
...
+	// Checking this first makes sure that we aren't running regex on everything
+	if strings.HasSuffix(name, `\`) {
+		suffix := regexp.MustCompile(`\\+$`).FindString(name)
+		if len(suffix)%2 == 1 {
+			name += `\`
+		}
+	}
+
 	util.RunCmd([]string{
 		"R",
```
dan-stowell, line 51: How does R know here to install the package?

ALANVF: Because that's where it installs to by default if it doesn't have write access to its normal library directory on linux (and it usually doesn't). For non-linux platforms (and well, it works on linux too), R can use the `$R_LIBS_USER` env var instead

dan-stowell: Got it.

PR #44, opened 2020-07-22
Fix R installation, search, specfile, and lockfile issues

ALANVF:
Any issues that R was having before have been fixed (for real this time).
Note to self: nullable types in Go are bad.


ALANVF:
That package didn't install because it requires R 3.5 or higher, but we only have R 3.4 on repl.it

ALANVF:
It should all be fixed now

PR #46, opened 2020-08-12
Allow python comments to hint UPM

Zabot:
UPM has difficulty resolving python packages when multiple packages provide the same module. Adds the ability to mark an import line with `#upm(package_name)` to override the package guess and directly specify the package that should be used.

```python
import twitter #upm package(python-twitter)

...
```

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -394,10 +394,17 @@ func guess(python string) (map[api.PkgName]bool, bool) {
 			continue
 		}
 
-		pkg, ok := moduleToPypiPackage[modname]
-		if ok {
+		if modname[0] == '"' && modname[len(modname) - 1] == '"' {
+			pkg := modname[1:len(modname) - 1]
+
 			name := api.PkgName(pkg)
 			pkgs[normalizePackageName(name)] = true
+		} else {
+			pkg, ok := moduleToPypiPackage[modname]
+			if ok {
+				name := api.PkgName(pkg)
+				pkgs[normalizePackageName(name)] = true
+			}
 		}
 	}
 
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -62,6 +62,12 @@ type poetryLock struct {
 	} `json:"package"`
 }
 
+// moduleMetadata represents the information that could be associated with
+// a module using a #upm pragma
+type modulePragmas struct {
+	Package string `json:"package"`
+}
+
 // normalizeSpec returns the version string from a Poetry spec, or the
 // empty string. The Poetry spec may be either a string or a
 // map[string]interface{} with a "version" key that is a string. If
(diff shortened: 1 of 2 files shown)


resources/python/pipreqs.py
```diff
@@ -51,8 +51,27 @@ def get_all_imports(
             with open_func(file_name, "r", encoding=encoding) as f:
                 contents = f.read()
             try:
+                # We need to be able to reference a pragma in the comments
+                lines = contents.split('\n')
                 tree = ast.parse(contents)
                 for node in ast.walk(tree):
+                    try:
```
turbio, line 58: doing this reconstruction on every single node in the AST sound expensive. would it be easy to reconstruct only when the `isinstance(node, ast.Import)` or `isinstance(node, ast.ImportFrom)` are hit. Then we can just tag the import with extra details from the pragma.

internal/backends/python/python.go
```diff
@@ -394,10 +394,17 @@ func guess(python string) (map[api.PkgName]bool, bool) {
 			continue
 		}
 
-		pkg, ok := moduleToPypiPackage[modname]
-		if ok {
+		if modname[0] == '"' && modname[len(modname) - 1] == '"' {
```
turbio, line 397: not sure i'm understanding this, what's the difference between a module with/without the surrounding quotes. Is this just to indicate the name is from a pragma?

Zabot: Yeah, the quotes are to indicate that the "module" is actually a package name extracted from a pragma, and that it should be used directly instead of being resolved though `moduleToPypiPackage`.

Otherwise we wouldn't be able to force upm to use a package with the same name as the module. e.g. We want to use the package `twitter` to provide the module `twitter` but `moduleToPypiPackage` resolves it to `python-twitter`. Adding a `#upm(twitter)` wouldn't help because the module name was already `twitter`, and `twitter` gets resolved to `python-twitter`.

resources/python/pipreqs.py
Zabot: I was thinking that the flexibility of being able to add a package that may not have an associated explicit import could be useful in optional dependency situations. I think your right though, the overhead of doing this and potentially catching an exception on every node in the AST is going to grow much faster then it did in my short tests. I'll tweak this!

internal/backends/python/python.go
dan-stowell: Here I come, barreling into code I've never read or written before, to ask if there's a more structured to indicate a package name comes from a pragma? Encoding it with quotes in the package name seems like it could lead to mistakes later on.

Zabot: We could encode everything into a JSON string and then change the parsing logic in the language backend. e.g.
```json
{
    "modules": {
        "twitter" : {
            "package": "python-twitter"
        },
        "module_with_no_pragma" : {}
    }
}
```
It increases the complexity a little, but it may be less error prone, and let us add other metadata if needed. I think I'll change the pragma to something like `#upm package(python-twitter)` to line up with the JSON structure.

Zabot: Somehow I totally missed that we were already serializing to JSON, I thought we were passing space delimitated strings. I'll definitely make the swap to a JSON format.

PR #47, opened 2020-08-12
Support passing project name to backend

Zabot:
Poetry cannot install packages with the same name as the project. UPM initializes the specfile automatically and does so with whatever the default project name is. Using poetry, the default project name is the name of the root directory.

This causes an error when the root directory of a project has the same name as a package that should be installed. This PR adds a `--name` flag that can be passed to upm to specify what to use as the project name in the specfile. If `--name` is not specified, the default project name is used as determined by the language backend.

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -248,18 +248,18 @@ func pythonMakeBackend(name string, python string) api.LanguageBackend {
 
 			return info
 		},
-		Init: func(name string) {
+		Add: func(pkgs map[api.PkgName]api.PkgSpec, projectName string) {
+			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
 				cmd := []string{python, "-m", "poetry", "init", "--no-interaction"}
 
-				if name != "" {
-					cmd = append(cmd, "--name", name)
+				if projectName != "" {
+					cmd = append(cmd, "--name", projectName)
 				}
 
 				util.RunCmd(cmd)
 			}
-		},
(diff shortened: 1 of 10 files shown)


internal/backends/python/python.go
```diff
@@ -248,10 +248,18 @@ func pythonMakeBackend(name string, python string) api.LanguageBackend {
...
+				cmd := []string{python, "-m", "poetry", "init", "--no-interaction"}
+
+				if name != "" {
+					cmd = append(cmd, "--name", name)
+				}
+
+				util.RunCmd(cmd)
 			}
+		},
+		Add: func(pkgs map[api.PkgName]api.PkgSpec) {
```
turbio, line 262: not sure but perhaps name should just be an arg to `Add` if it's not really used elsewhere.

Zabot: My thought process was that the name of the project isn't strictly relevant to adding a package, its only meaningful in initializing a project, which as of right now we implicitly do when we add the first package. If we need to pass other initialization specific information, e.g. interpreter path, package repositories, etc, this interface could get cluttered very quickly. 

I think this and the other comment both get at if initializing a specfile should be an implicit part of adding a package, or if it should happen independently. Not as a separate command `upm add ...` should still call `backend.Init` and then `backend.Add`, but the two operations would be logically decoupled.

dan-stowell: I hear your reasoning, @Zabot. We've tended to solve the problem right in front of us with notes on what to do in the future. If `upm add` is the only spot where `upm` will create a spec file, then let's pass the name as an argument to `Add` and leave a nice comment saying why we did that.

PR #48, opened 2020-08-19
Generate python package mapping

Zabot:
The `pypi_package.json` mapping required several separate internal tools to generate a usable package list for mapping python modules to packages.

This PR creates a single tool that fetches the python package index, determines the modules in each package, and generates the corresponding go source. It also generates an artifact `cache.json` that can be updated iteratively by future executions, only requiring analysis of new packages, as determined by comparing the sha of the most recent release with the sha of the cached data.


ryantm:
Somehow we generate the package mapping as is, so I think this is outdated.

PR #51, opened 2020-09-01
Lazily load python package mappings

cbrewster:
These maps were always being loaded into memory on startup, causing significant overhead for every UPM command.

Before:
Messages Image(1355001053)

After:
Screen Shot 2020-09-01 at 11 18 07 AM


cbrewster:
Yeah I was pretty surprised when I saw the map show up in the profile when running `upm help`, must be something with go's initialization model.

amasad:
nice

PR #52, opened 2020-09-10
Add Ruby guessing (only for preinstalled Polygott gems)

cbrewster:
This adds very basic Ruby gem guessing by walking the Ruby AST and looking for any `require 'somegem'` nodes and checking against a list of allowed guessable gems.

As noted in the guessing Ruby script, there isn't always a 1-to-1 mapping between what you `require` and the name of the gem to install. We want to move to using bundler to run Ruby repls without breaking compatibility with preexisting Ruby repls. When running via Bundler, globally installed gems are excluded from the environment. This means we need to, at least, guess the [preinstalled Polygott gems](https://github.com/replit/polygott/blob/7721bd5c9c8c50013801d1c581fb94a4508b2128/languages/ruby.toml#L16) (I excluded `rufo`, `rspec` and `solargraph` as those are used as binaries).

Before merging, we need to add the `parser` gem to Polygott.

### DIFF
--- a/resources/ruby/guess-gems.rb
+++ b/resources/ruby/guess-gems.rb
@@ -1,0 +1,60 @@
+require 'parser/current'
+require 'json'
+
+# For now, we will only guess gems that come preinstalled in Polygott.
+# The mapping between what you require and the gem name is not always 1-to-1,
+# so we will need some sort of mapping like what we do with PyPi if we want
+# to do complete ruby gem guessing.
+$allowed_gems = [
+  "sinatra",
+  "stripe"
+]
+
+def guess_gems(code, guesses)
+  root = Parser::CurrentRuby.parse(code)
+  traverse_node(root, guesses)
+end
+
+def traverse_node(node, guesses)
+  if node.class != Parser::AST::Node
+    return
+  end
+  
+  # Looking for any s(:send, nil, :require, s(:str, "gem"))
+  if node.type == :send && node.children[1] == :require
+    req_node = node.children[2]
+    if req_node.class == Parser::AST::Node && req_node.type == :str
+      process_require(req_node.children.first, guesses)
+      return
+    end
+  end
+
+  node.children.each { |child|
+    traverse_node(child, guesses)
+  }
+end
+
+def process_require(req_str, guesses)
+  if req_str.empty?
+    return
+  end
+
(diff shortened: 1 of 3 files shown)


cbrewster:
Looking at https://guides.rubygems.org/name-your-gem/ we might be able to follow those conventions when guessing gems.

cbrewster:
**For future reference if we want to add full Ruby gem guessing support:**
It looks like we could generate a mapping between RubyGem names and require names following this convention: https://guides.rubygems.org/name-your-gem/
We can't blind follow the convention from require -> gem name because of cases like:
`require 'sinatra/cookies'` coming from `sinatra-contrib`
But we can do something similar to Python where we generate a mapping from require to RubyGem name based on all the gems on RubyGems.

resources/ruby/guess-gems.rb
```diff
@@ -0,0 +1,60 @@
...
+  if req_str.start_with?('/') || req_str.start_with?('.')
+    return
+  end
+
+  # Handle requires like 'sinatra' or 'sinatra/content_for'
+  gem = req_str.split('/')[0]
+  if $allowed_gems.include?(gem)
+    guesses[gem] = true
+  end
+end
```
kochman, line 52: Is this too broad? I'm wondering if another import with the "sinatra/" prefix might not necessarily be part of the "sinatra" gem, based on what you wrote in your comment about require names not mapping 1:1 to gem names.

kochman (approved):
LGTM! Tested it out locally, it got sinatra and stripe and ignored others.

resources/ruby/guess-gems.rb
cbrewster: Yeah, I'm somewhat torn on this one. For example, this script uses the `parser` gem which I require via `require 'parser/current'`. There's also `sinatra/base` which comes from the `sinatra` gem, but on the flip-side there iss `sinatra/namespace` which comes from `sinatra-contrib`.

cbrewster: Actually, I'll just make the script look for `sinatra` and `sinatra/base`, but ignore any other `sinatra/<things>`

kochman: Cool, being more conservative here seems like the safe move 👍

PR #56, opened 2020-11-13
Use PyPI REST API for fetching Python package info

sergeichestakov:
`upm info` is broken for Python because we're still relying on PyPI's XML-RPC API which is getting deprecated in the near future and is currently being ratelimited very heavily. From [the docs](https://warehouse.readthedocs.io/api-reference/xml-rpc.html):
> The XML-RPC API will be deprecated in the future. Use of this API is not recommended, and existing consumers of the API should migrate to the RSS and/or JSON APIs instead.
As a result, this API has a very restrictive rate limit and it may be necessary to pause between successive requests.

This PR rewrites the pypi-info script (which is what actually fetches the info about a given Python package) to use PyPI's REST API (API reference [here](https://warehouse.readthedocs.io/api-reference/json.html)) and updates the Go types to match since some of the fields are slightly different (most notably, project_url is a string instead of a string[] and the bug tracking, documentation, and package urls are their own fields).

Note that `pypi-search` still relies on the legacy XML-RPC API but it looks like it'll have to keep doing so for the time being as the REST API does not expose an endpoint for search (there are only two endpoints: `/pypi/<project_name>/json` and `/pypi/<project_name>/<version>/json`). Will need to look into alternatives there and follow up, but search seems to still be working for now.


resources/python/pypi-info.py
```diff
@@ -1,24 +1,23 @@
...
-# the XMLRPC API. It takes one argument, the name of the package (not
+# This is a Python script that looks up package metadata on PyPI.
+# It takes one argument, the name of the package (not
 # necessarily canonical), and outputs the results in JSON format (a
-# map, see pypiXMLRPCInfo). The script works on both Python 2 and
-# Python 3.
+# map). The script works on both Python 2 and Python 3.
 
 from __future__ import print_function
+import requests
```
masad-frost, line 7: `requests` is not part of the standard library and so this script won't work on all systems. You can use `urllib.request` or IMO we shouldn't even be shelling out for this, let's do the request from within the upm process in Go

internal/backends/python/python.go
```diff
@@ -190,57 +192,24 @@ func pythonMakeBackend(name string, python string) api.LanguageBackend {
...
-				Description: output.Summary,
-				Version:     output.Version,
-				HomepageURL: output.HomePage,
+				Name:             output.Name,
+				Description:      output.Summary,
+				Version:          output.Version,
+				HomepageURL:      output.HomePage,
+				DocumentationURL: output.DocsURL,
+				BugTrackerURL:    output.BugTrackerURL,
 				Author: util.AuthorInfo{
```
masad-frost, line 207: Sweet!

masad-frost (dismissed):
The script won't work on systems where `requests` is not installed in the global `sitepackages`

resources/python/pypi-info.py
masad-frost: @turbio is there a good reason you can think of why we're shelling out here?

masad-frost (commented):
[redacted]

resources/python/pypi-info.py
turbio: yeah, this seems like something we should just do in go land :shrug: 
I think it was original python since they've got a nice xmlrpc api

sergeichestakov: But I like Python :(

sergeichestakov: Fixed

internal/backends/python/python.go
```diff
@@ -185,65 +196,44 @@ func pythonMakeBackend(name string, python string) api.LanguageBackend {
...
-				matched, err := regexp.MatchString(`(?i)doc`, name)
-				if err != nil {
-					panic(err)
-				}
-				if matched {
-					info.DocumentationURL = url
-					continue
-				}
+			if res.StatusCode != 200 {
+				return api.PkgInfo{}
```
turbio, line 208: is this what we used to do? Silently failing and returning garbage isn't normaly the move.

turbio (changes requested):

internal/backends/python/python.go
sergeichestakov: yeah this matches the current behavior (at least for 404s): https://github.com/replit/upm/blob/4caf33404424a0199db79615a64eb02011e8e93b/resources/python/pypi-info.py#L19-L20

Returning an empty PkgInfo struct will result in a message printing "no such package": https://github.com/replit/upm/blob/2e4a14f6d8d596a178dcf3eb75fc21498e869423/internal/cli/cmds.go#L99-L101

sergeichestakov: updated to only return an empty struct on 404s and die otherwise

sergeichestakov: seems to be a standard pattern in the other language backends: https://github.com/replit/upm/blob/a3fb24a65e201bd582303ec0d292ec9f4bf6c0f8/internal/backends/java/java.go#L281-L283

turbio (approved):
lgtm :snake: :package: :snake:

PR #59, opened 2020-12-17
Search python packages from local cache

Zabot:
PyPi is deprecating the XMLRPC search interface. This searches for packages in the local UPM package mapping and displays results in order of downloads.


dan-stowell (approved):
🩹

PR #60, opened 2020-12-23
Fix nodejs backend not catching backtick only requires

tjk:
This was a funky bug to catch! As soon as we included a require without backticks, the backtick require worked.


cdmistman:
closing and rolling into #109 - i didn't realize backticks are valid for requires, fixing there...

PR #64, opened 2021-05-07
Ignore nodejs requires prefixed with 'node:'

cbrewster:
Asana: https://app.asana.com/0/1199976029439378/1200303425700772

TL;DR: In Node.js 16 you can prefix core module requires with `node:`
Ex:
```
// old
const fs = require("fs");
// new
const fs = require("node:fs");
```

This breaks guessing because our internal module list only looks for `fs` and not `node:fs`.
Solution: Skip any modules with the `node:` prefix when guessing imports. 

## Test
Manually tested that guessing does not include `node:` prefixed modules.


lhchavez (approved):
TIL about this magical prefix!

PR #66, opened 2021-06-07
Fixed the support for http/https imports in nodejs

Minivera:
We would parse http/https imports as NPM import through UPM, which would cause NPM to trigger errors as it tries to import packages using the URL path. Nodejs is on its way to support URL import paths and the browser already does it (Which is used by CDNs and tools like skypack), so this change should allow support for both future version of node and imports for the browser.

This is a relatively small change, so I took the opportunity to add a test suite for this backend and a make target to run the tests on demand.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,3 +65,8 @@ help: ## Show this message
 		sed 's/:[^#]*[#]# /|/'		| \
 		sed 's/%/LANG/'			| \
 		column -t -s'|' >&2
+
+.PHONY: test
+test: ## Run the tests
+	make upm
+	go test ./... -v
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,11 @@ export GO111MODULE=on
 .PHONY: upm
 upm: cmd/upm/upm ## Build the UPM binary
 
-cmd/upm/upm: $(SOURCES) $(RESOURCES) $(GENERATED)
-	go run github.com/rakyll/statik -src resources -dest internal -f
+cmd/upm/upm: $(SOURCES) $(RESOURCES) $(GENERATED) statik
 	cd cmd/upm && go build -ldflags "-X 'github.com/replit/upm/internal/cli.version=$${VERSION:-development version}'"
+
+statik:
+	go run github.com/rakyll/statik -src resources -dest internal -f
 
 internal/backends/python/pypi_map.gen.go: internal/backends/python/pypi_packages.json
 	go generate ./internal/backends/python
@@ -67,6 +69,6 @@ help: ## Show this message
 		column -t -s'|' >&2
 
 .PHONY: test
-test: ## Run the tests
+test: statik ## Run the tests
 	make upm
 	go test ./... -v
--- a/Makefile
+++ b/Makefile
@@ -70,5 +70,4 @@ help: ## Show this message
 
 .PHONY: test
 test: statik ## Run the tests
-	make upm
 	go test ./... -v
--- a/internal/backends/nodejs/nodejs_test.go
+++ b/internal/backends/nodejs/nodejs_test.go
@@ -1,0 +1,112 @@
+package nodejs
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/replit/upm/internal/api"
+)
+
+func TestNodejsYarnBackend_Guess(t *testing.T) {
+	tcs := []struct {
+		scenario    string
+		backend     api.LanguageBackend
+		fileContent string
+		expected    map[api.PkgName]bool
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -65,3 +65,8 @@ help: ## Show this message
 		sed 's/:[^#]*[#]# /|/'		| \
 		sed 's/%/LANG/'			| \
 		column -t -s'|' >&2
+
+.PHONY: test
+test: ## Run the tests
+	make upm
```
cbrewster, line 71:
```suggestion
```
Do we need to build upm to run the tests?

internal/backends/nodejs/nodejs_test.go
```diff
@@ -0,0 +1,112 @@
...
+		`,
+			expected: map[api.PkgName]bool{
+				"react":             true,
+				"@material-ui/core": true,
+			},
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.scenario, func(t *testing.T) {
```
cbrewster, line 79:
```suggestion
	for _, tc := range tcs {
		tc := tc
		t.Run(tc.scenario, func(t *testing.T) {
```

This probably doesn't matter too much since the subtest aren't marked as parallel. But in case we do mark it as parallel at some point it might be nice to do this to save us from https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

internal/backends/nodejs/nodejs_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/replit/upm/internal/api"
+)
+
+func TestNodejsYarnBackend_Guess(t *testing.T) {
```
cbrewster, line 11: :heart_eyes:

cbrewster (approved):
:globe_with_meridians:

Makefile
Minivera: We need the `statik` folder to run the tests, otherwise it won't compile. I'm not the biggest make expert, would copying the `go run github.com/rakyll/statik -src resources -dest internal -f` line from `make upm` do the same thing?

cbrewster: Ah I see, we could probably split that out to its own make target and have both `cmd/upm/upm` and `test` depend on it.

For now we could also do:
```go
.PHONY: test
test: cmd/upm/upm ## Run the tests
    go test ./... -v
```
This will do the exact same thing as what you have but in a slightly more idiomatic way

Minivera: Having the statik command as a separate target and making it a dep of both build and test seems to work :smile:

Makefile
```diff
@@ -7,10 +7,12 @@ export GO111MODULE=on
 .PHONY: upm
 upm: cmd/upm/upm ## Build the UPM binary
 
-cmd/upm/upm: $(SOURCES) $(RESOURCES) $(GENERATED)
-	go run github.com/rakyll/statik -src resources -dest internal -f
+cmd/upm/upm: $(SOURCES) $(RESOURCES) $(GENERATED) statik
 	cd cmd/upm && go build -ldflags "-X 'github.com/replit/upm/internal/cli.version=$${VERSION:-development version}'"
 
+statik:
```
lhchavez, line 13: small nit: Make depends on the left-hand-side of every rule be a path that is generated by the rule. if it does not create it, this rule will always be "out of date", and successive `make` invocations will always try to run this:

```shell
lhchavez@luisval:~/upm$ make cmd/upm/upm
go run github.com/rakyll/statik -src resources -dest internal -f
cd cmd/upm && go build -ldflags "-X 'github.com/replit/upm/internal/cli.version=${VERSION:-development version}'"
lhchavez@luisval:~/upm$ make cmd/upm/upm
go run github.com/rakyll/statik -src resources -dest internal -f
cd cmd/upm && go build -ldflags "-X 'github.com/replit/upm/internal/cli.version=${VERSION:-development version}'"
lhchavez@luisval:~/upm$ make cmd/upm/upm
go run github.com/rakyll/statik -src resources -dest internal -f
cd cmd/upm && go build -ldflags "-X 'github.com/replit/upm/internal/cli.version=${VERSION:-development version}'"
```

_so_ the correct way to express this is the following:

```suggestion
internal/statik/statik.go: $(shell find resources -type f)
```

and have L10 and L72 depend on `internal/statik/statik.go` instead of `statik`. the `$(shell ...)` expression also has the effect of forcing this rule to be marked as stale if any of the files in `resources/` change.

Minivera: :+1: Will create a follow up PR

PR #67, opened 2021-06-07
Implemented a small change to the makefile

Minivera:
This change should make sure that the statik rule won't rerun every time the build or test targets are executed. This should save recreating the file only when its dependencies change.


lhchavez (approved):
:flight_departure:

PR #70, opened 2021-06-16
Add nix build files

Minivera:
This PR adds the files from https://github.com/replit/nixpkgs-replit/tree/main/pkgs/upm directly to upm to reduce the number of repositories we have to update to push changes.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -8,9 +8,7 @@ in
 let
     src = ./.;
 
-    gitSrc = builtins.filterSource
-                   (path: type: true)
-                   ./.;
+    gitSrc = pkgs.copyPathToStore ./.;
     revision = runCommand "get-rev" {
         nativeBuildInputs = [ git ];
         dummy = builtins.currentTime;
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,22 @@
+{ lib, buildGoModule, fetchFromGitHub, statik }:
+
+buildGoModule rec {
+    name = "upm";
+    version = "908674b";
+
+    src = fetchFromGitHub{
+        owner = "replit";
+        repo = "upm";
+        rev = "${version}";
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -0,0 +1,22 @@
...
+buildGoModule rec {
+    name = "upm";
+    version = "908674b";
+
+    src = fetchFromGitHub{
+        owner = "replit";
+        repo = "upm";
+        rev = "${version}";
+        sha256 = "0y50q2fxz1q1xbxirjka70y247x70xxz8kj1h8dmgrzd793k3y6s";
+    };
```
cbrewster, line 12:
```suggestion
    src = ./.;
```

default.nix
```diff
@@ -0,0 +1,22 @@
+{ lib, buildGoModule, fetchFromGitHub, statik }:
+
+buildGoModule rec {
+    name = "upm";
+    version = "908674b";
```
cbrewster, line 5: Either we can pass the version in as a parameter, or we can omit it entirely. It's primarily for human consumption.

Minivera:
@cbrewster Letting you know I pushed all the changes you requested in prybar here as well

default.nix
```diff
@@ -0,0 +1,28 @@
+{ pkgs ? import <nixpkgs>{} } :
+let
+    buildGoModule = pkgs.buildGoModule;
```
cbrewster, line 3: Can we move these inline as well?

cbrewster (approved):
:snowflake:

default.nix
Minivera: I could convert this part of the code to something like this: 
```
let
    inherit(pkgs)
        buildGoModule
        statik
        git
        runCommand;
in
```

Would that work better?

cbrewster: Yeah either works, I suppose it's mainly stylistic haha. Up to you

PR #71, opened 2021-06-21
Add support for fetching the custom version through the API

Minivera:
When the dynamic version code was executed through niv, the .git folder would be missing. This is because niv downloads the archive from github using an URL without getting the .git folder, since it is never really pushed.

That's a big problem for us and this PR fixes that issue by adding an if and checking if said .git folder exists. If not, it will use the version information from the API and parse the short sha from there.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -4,17 +4,28 @@ let
         buildGoModule
         statik
         git
+        python3
         runCommand;
 in
 let
+    packageName = "upm";
+
     src = pkgs.copyPathToStore ./.;
+    versionFile = builtins.fetchurl https://api.github.com/repos/replit/upm/commits/master;
     revision = runCommand "get-rev" {
-        nativeBuildInputs = [ git ];
+        nativeBuildInputs = [ git python3 ];
         # impure, do every time, see https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/fetchgitlocal/default.nix#L9
         dummy = builtins.currentTime;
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -4,15 +4,24 @@ let
         buildGoModule
         statik
         git
+        python3
         runCommand;
 in
 let
     src = pkgs.copyPathToStore ./.;
+    versionFile = builtins.fetchurl https://api.github.com/repos/replit/upm/commits/master;
```
cbrewster, line 12: This will always pull the latest SHA right? This may give us incorrect version tagging if upm/master is ahead of the commit we have pinned in goval.

Maybe we can make version be an argument and then we can pass it in when we import the package in the pid1 derivation (since the fetchgit result should contain the SHA)

Minivera: Yeah, that could work too. That way we get the best of both worlds, though I wonder if we even need a dynamic version if we're always going to fetch the pinned sha from github :thinking:

cbrewster: Indeed, the main reason we want any of this is to know what version of the binaries we have in prod. As long as we know the git SHA for pid1, that lets us trace all the versions of its dependencies since we pin them via niv.

Minivera: I can switch to a parameter based versionName in the bash if for now, dynamic versioning would be nice to have if we ever want to build and publish the derivations from CI.

Minivera:
@cbrewster I implemented the changes you requested, anything else you'd like to see improve? :slightly_smiling_face:

cbrewster (approved):
LGTM!

PR #73, opened 2021-06-28
Implement Nix in CircleCI

Minivera:
This flow currently only builds the nix package using the official Nix docker images. It will run all phases, but will not test the package once built. It does run the checks though.

### DIFF
--- a/result
+++ b/result
@@ -1,1 +1,0 @@
-/nix/store/rad9dm9x4j6l8vzpw7wal04jpdsj2fl7-upm-ac649e7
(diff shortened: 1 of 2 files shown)


result
```diff
@@ -1 +0,0 @@
-/nix/store/rad9dm9x4j6l8vzpw7wal04jpdsj2fl7-upm-ac649e7
```
Minivera, line 1: Not sure why this was pushed, it definitely shouldn't have

lhchavez (approved):
:green_circle: 

if it passes on CI, should be good.

PR #74, opened 2021-07-05
Use Nix flake

cbrewster:
This adds a `flake.nix` file which locks nixpkgs and uses the fancy new Nix 2.0 commands. It also has built-in support for getting the git revision for us. The default.nix file provides a flake-compat shim which allows upm to be built using the old Nix commands like `nix-build`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,7 @@ jobs:
       - checkout
       - restore_cache:
           key: nix-store
+      - run: nix-env -iA nixpkgs.git
       - run: nix-build
       - save_cache:
           key: nix-store
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,8 @@ jobs:
       - restore_cache:
           key: nix-store
       - run: nix-env -iA nixpkgs.git
+      - run: git diff
+      - run: git status
       - run: nix-build
       - save_cache:
           key: nix-store
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,6 @@
 *.gen.go
 /.vscode/
 /.idea/
+
+# nix build creates a "result" symlink to the nix store output
 result
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 *.gen.go
 /.vscode/
 /.idea/
+result
--- a/flake.nix
+++ b/flake.nix
@@ -11,7 +11,7 @@
       (system:
         let
           pkgs = nixpkgs.legacyPackages.${system};
-          upm = pkgs.callPackage ./default.nix {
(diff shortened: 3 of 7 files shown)


.circleci/config.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       - checkout
       - restore_cache:
           key: nix-store
+      - run: nix-env -iA nixpkgs.git
```
Minivera, line 24: Do we want to still try to `nix-build` as usual to see f this is also compatible outside nix flakes?

result
```diff
@@ -0,0 +1 @@
+/nix/store/63pln8qafch9hhjfvp4lm6hha5db7vd3-upm-0.1.0-50e80cc
```
Minivera, line 1: This should probably not be added to the store. Maybe we need to add it to the `.gitignore`? :thinking:

flake.nix
```diff
@@ -0,0 +1,25 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils }:
+    flake-utils.lib.eachDefaultSystem
+      (system:
+        let
+          pkgs = nixpkgs.legacyPackages.${system};
+          upm = pkgs.callPackage ./upm.nix {
+            rev = if self ? rev then "0.1.0-${builtins.substring 0 7 self.rev}" else "0.1.0-dirty";
```
Minivera, line 15:
```suggestion
            rev = if self ? rev then "0.0.0-${builtins.substring 0 7 self.rev}" else "0.0.0-dirty";
```
Small nitpick, we could try keeping this as "version 0" to clearly state that our versions depend on the revision and not semantic versioning? I will be creating an issue on the nix repo about this hard "must start with a digit" requirement.

.circleci/config.yml
cbrewster: Yup, this was just because CI was complaining about `git` not being present.

result
cbrewster: Oops!

flake.nix
cbrewster: Sure thing!

lhchavez (approved):
:snowman:

.gitignore
```diff
@@ -8,3 +8,4 @@
 *.gen.go
 /.vscode/
 /.idea/
+result
```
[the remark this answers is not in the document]
cbrewster: Yeah I'll add a comment. This happens with and without flakes 😞 
I guess the normal Nix development flow would be like: `nix-build && ./result/bin/my_cool_bin`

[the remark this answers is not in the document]
Minivera: I pushed it a few times in prybar and upm by mistake as well

PR #87, opened 2022-10-13
Disable esparse's TS option to avoid removal of unused imports

airportyh:
# Why?

We have a bug where if you first run a file with an un-used import, say:

```ts
import isOdd from 'is-odd';
```

Then you update the code to use the import and run it again:

```ts
import isOdd from 'is-odd';
console.log(isOdd(3));
```

You get a module 'is-odd' is not found error because:

1. upm failed to auto install the module during the 1st run because the unused import was striped by [esparse (fork of esbuild)](https://github.com/amasad/esparse) before the guesser could include the module
2. in the second run, the guesser used the cached result from the first run because the hash key is based on import statements grabbed using regexp searches instead of eparse: https://github.com/replit/upm/blob/master/internal/store/util.go#L32

More context:
https://app.asana.com/0/1201647179266797/1202932490289607/f

# What Changed?

1. disabled the TS.Parse option of esparse
2. disabled the MangleSyntax option of esparse
3. the above 2 together disable the removal of unused imports in the parsing process: https://github.com/amasad/esparse/blob/master/parser/parser.go#L7614. Basically we put the parser into JS parse mode that can cope with TS but just ignores the semantics of TS. This is ok for what we want: just the list of import statements. This also has the nice effect that the user can write the statement: `import something from 'something'`, run it, and get the library installed, which I think is more intuitive.

### DIFF
--- a/internal/backends/nodejs/grab.go
+++ b/internal/backends/nodejs/grab.go
@@ -1,15 +1,16 @@
 package nodejs
 
 import (
+	"io/ioutil"
+	"log"
+	"path/filepath"
+	"strings"
+
 	"github.com/amasad/esparse/ast"
 	"github.com/amasad/esparse/logging"
 	"github.com/amasad/esparse/parser"
 	"github.com/replit/upm/internal/api"
 	"github.com/replit/upm/internal/util"
-	"io/ioutil"
-	"log"
-	"path/filepath"
-	"strings"
 )
 
 var internalModules = []string{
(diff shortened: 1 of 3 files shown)


gburtini (approved):
@airportyh and I talked offline. He's going to add a few test cases here, one for a hard-to-parse TypeScript case and another for to catch this regression. 

[For posterity, this was originally introduced here](https://github.com/replit/upm/commit/0be80ecbd18f0d9a74d96158ffce2c8c1a6a87a0) and includes a comment about the speed (cost?) of React parsing. There are some existing [tests on React-ish JS over here](https://github.com/replit/upm/blob/master/internal/backends/nodejs/nodejs_test.go) that we also considered.

airportyh:
> @airportyh and I talked offline. He's going to add a few test cases here, one for a hard-to-parse TypeScript case and another for to catch this regression.

Thanks! Turns out the original tests failed the moment I ran the original testcases as .ts files. Conditional import/requires don't work if they come after typings.

PR #91, opened 2022-12-17
(nodejs): Catch internal submodules (ex. dns/promises)

hackermondev:
Right now, UPM doesn't catch internal submodules correctly.

For example:
```js
require('dns/promises')
```

will cause to UPM to try and install the ``dns`` package even though it's an internal module.

@masad-frost

### DIFF
--- a/internal/backends/nodejs/grab.go
+++ b/internal/backends/nodejs/grab.go
@@ -1,7 +1,6 @@
 package nodejs
 
 import (
-	"fmt"
 	"io/ioutil"
 	"log"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


internal/backends/nodejs/grab.go
```diff
@@ -163,7 +164,7 @@ func guessBareImports() map[api.PkgName]bool {
 
 			isInternalMod := false
 			for _, internal := range internalModules {
-				if internal == mod {
+				if internal == mod || strings.HasPrefix(mod, fmt.Sprintf("%s/", internal)) {
```
masad-frost, line 167: I wonder if it's better if it's better if we moved lines 201->211 to be above this.

masad-frost (approved):
Thank you!

PR #94, opened 2023-02-27
Fix maven API requests using http instead of https

masad-frost:
Looks like we use http instead of https when talking to maven and that suddenly stopped working.

Tests in msch_test.go should have us covered.


ryantm (approved):
lgtm

masad-frost:
oops didn't see your PR

PR #95, opened 2023-02-28
rlang: use https for search

ryantm:
Why
===

* We recently had trouble with Maven becoming inaccessible via http
* Let's preemptively change this last one.

What Changed
============

* Changed URL to https

Test plan
=========
* `nix run .# -- search -l rlang limit`


masad-frost (approved):
:lock:

PR #102, opened 2023-03-30
Use Semaphore for CI

cdmistman:


cdmistman:
semaphore works, i asked for circleci to be removed so once it's removed we won't get the ugly `x`s anymore :)

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,123 @@
...
+            - cache restore goreleaser
+            - goreleaser --release-notes=<(scripts/extract-changelog.py)
+        - name: snapcraft
+          commands:
+            - >-
+              snapcraft login --with <(
+                echo "$SNAPCRAFT_LOGIN_FILE" |
+                base64 --decode --ignore-garbage
+              )
+            - scripts/upload-snap.bash
```
cdmistman, line 113: haven't tested, but if we ever do another release we can figure that out then :) things are mostly implemented

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,123 @@
...
+        - name: nix
+          commands:
+            - cache restore nix-store
+            - nix build
+            - ./result/bin/upm --version
+            - cache store nix-store /nix
+        # TODO: get this working again
+        # - name: docker
+        #   commands:
+        #     - make light full
```
cdmistman, line 71: some package we depend on in the docker images doesn't have the right version when building the image, i figured it wasn't important enough to keep diagnosing

masad-frost: yeah let's skip the docker builds

.semaphore/semaphore.yml
masad-frost, line 113: I'd also skip snap

masad-frost (approved):

PR #107, opened 2023-04-26
Call poetry as an executable instead of as a module of python

airportyh:
# Why?

Needed for the Python Nix module to work. With the poetry setup where it is installed in its own virtual environment, we can no longer put poetry on the library path of python and therefore we cannot call poetry as a module. Instead we need to call it as an executable.

## Changes

calling poetry as a standalone executable

## Testing

1. compile upm via `make upm; make install`
2. upload upm to a repl
3. hack the `PATH` to make it show up first
4. activate the python module `modules = ["python"]`
5. setup a `PYTHONUSERBASE` variable as `$HOME/$REPL_SLUG/.pythonlibs`
6. `poetry init` in the repl directory if this was a blank repl
7. try using the packager to install a python package like pyyaml

Play in this repl which already has the above set up: https://replit.com/join/vvfytgyghu-tobyho

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -191,7 +191,7 @@ func pythonMakeBackend(name string, python string) api.LanguageBackend {
 			// so complicated??)
 
 			outputB := util.GetCmdOutput([]string{
-				python, "-m", "poetry",
+				"poetry",
 				"config", "settings.virtualenvs.path",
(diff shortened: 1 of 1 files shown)


ryantm (approved):
I confirmed this doesn't break the regular Python template package installation experience.

PR #108, opened 2023-05-30
Add PNPM support

DillonB07:
This PR adds support for the NodeJS package manager [PNPM](https://pnpm.io).

I have tested this by running the command specified in the README and there were no issues.

One thing to note is running `pnpm init` will force the user to interact with no flag to accept all the defaults. Because of this, I have used `npm init -y` for creating the `package.json` file.

### DIFF
--- a/internal/backends/nodejs/nodejs.go
+++ b/internal/backends/nodejs/nodejs.go
@@ -1,4 +1,4 @@
-// Package nodejs provides backends for Node.js using Yarn and NPM.
+// Package nodejs provides backends for Node.js using Yarn, PNPM and NPM.
 package nodejs
 
 import (
@@ -81,7 +81,7 @@ type packageLockJSON struct {
 // nodejsPatterns is the FilenamePatterns value for NodejsBackend.
 var nodejsPatterns = []string{"*.js", "*.ts", "*.jsx", "*.tsx", "*.mjs", "*.cjs"}
 
-// nodejsSearch implements Search for nodejs-yarn and nodejs-npm.
+// nodejsSearch implements Search for nodejs-yarn, nodejs-pnpm and nodejs-npm.
 func nodejsSearch(query string) []api.PkgInfo {
 	// Special case: if search query is only one character, the
 	// API doesn't return any results. The web interface to NPM
@@ -135,7 +135,7 @@ func nodejsSearch(query string) []api.PkgInfo {
 	return results
 }
 
-// nodejsInfo implements Info for nodejs-yarn and nodejs-npm.
+// nodejsInfo implements Info for nodejs-yarn, nodejs-pnpm and nodejs-npm.
 func nodejsInfo(name api.PkgName) api.PkgInfo {
 	endpoint := "https://registry.npmjs.org"
 	path := "/" + url.QueryEscape(string(name))
@@ -203,7 +203,7 @@ func nodejsInfo(name api.PkgName) api.PkgInfo {
 	}
 }
 
-// nodejsListSpecfile implements ListSpecfile for nodejs-yarn and
+// nodejsListSpecfile implements ListSpecfile for nodejs-yarn, nodejs-pnpm and
 // nodejs-npm.
 func nodejsListSpecfile() map[api.PkgName]api.PkgSpec {
 	contentsB, err := ioutil.ReadFile("package.json")
@@ -224,7 +224,7 @@ func nodejsListSpecfile() map[api.PkgName]api.PkgSpec {
 	return pkgs
 }
 
-// nodejsGuessRegexps is the value of GuessRegexps for nodejs-yarn and
+// nodejsGuessRegexps is the value of GuessRegexps for nodejs-yarn, nodejs-pnpm and
 // nodejs-npm.
 var nodejsGuessRegexps = util.Regexps([]string{
 	// import defaultExport from "module-name";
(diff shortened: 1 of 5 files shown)


masad-frost:
Nice, thanks.

internal/backends/nodejs/nodejs.go
```diff
@@ -309,6 +309,67 @@ var NodejsYarnBackend = api.LanguageBackend{
...
+		api.QuirksAddRemoveAlsoInstalls |
+		api.QuirksLockAlsoInstalls,
+	GetPackageDir: func() string {
+		return "node_modules"
+	},
+	Search: nodejsSearch,
+	Info:   nodejsInfo,
+	Add: func(pkgs map[api.PkgName]api.PkgSpec, projectName string) {
+		if !util.Exists("package.json") {
+			util.RunCmd([]string{"npm", "init", "-y"})
```
masad-frost, line 327: Why not `pnpm init`?

internal/backends/nodejs/nodejs.go
```diff
@@ -309,6 +309,67 @@ var NodejsYarnBackend = api.LanguageBackend{
...
+	Remove: func(pkgs map[api.PkgName]bool) {
+		cmd := []string{"pnpm", "remove"}
+		for name := range pkgs {
+			cmd = append(cmd, string(name))
+		}
+		util.RunCmd(cmd)
+	},
+	Lock: func() {
+		util.RunCmd([]string{"pnpm", "install"})
+	},
```
masad-frost, line 348: Maybe we can `pnpm install --lockfile-only` and remove the LockAlsoInstalls quirk?

masad-frost: Actually this quirk is a good thing since it cuts extra calls to the package manager. Let's keep it as it is.

internal/backends/nodejs/nodejs.go
```diff
@@ -309,6 +309,67 @@ var NodejsYarnBackend = api.LanguageBackend{
...
+		}
+		contents := string(contentsB)
+		r := regexp.MustCompile(`(?m)^"?([^@ \n]+).+:\n  version "(.+)"`)
+		pkgs := map[api.PkgName]api.PkgVersion{}
+		for _, match := range r.FindAllStringSubmatch(contents, -1) {
+			name := api.PkgName(match[1])
+			version := api.PkgVersion(match[2])
+			pkgs[name] = version
+		}
+		return pkgs
```
masad-frost, line 366: Have you considered using `pnpm list --json --depth 0`, instead? I feel like reading lock files and parsing them is error prone and bordering the internals of package manager

masad-frost: I guess this is what we do for all other, cc @cdmistman we should consider using commands provided by package managers if they provide a machine readable format

masad-frost (approved):
Looks good to me. 

Ideally we write tests for these backends similar to https://github.com/replit/upm/blob/main/internal/backends/php/php_test.go and https://github.com/replit/upm/blob/main/internal/backends/rust/rust_test.go

I noticed we don't have any for our javascript backends :cry:. Do you mind writing them for `pnpm`?

@cdmistman can you see this to prod with your tree-sitter changes?

internal/backends/nodejs/nodejs.go
cdmistman: yeah, this should be `pnpm init` - we should only use the tools that are being used in the environment. shouldn't rely on `npm` being available if we're only promised `pnpm`

internal/backends/nodejs/nodejs.go
cdmistman: agreed, `LockAlsoInstalls` is good quirk :)

internal/backends/nodejs/nodejs.go
cdmistman: agreed that we shouldn't rely on lockfile. if we do, then it's definitely preferred to use the appropriate de/serialization for it - for example, this code shouldn't use a regex, it should instead deserialize the lockfile and read the entries via the deserialized map.

i'm not convinced that shelling out to `pnpm` is the best solution here. it has to load up node, which is big and heavy - we're already concerned about performance in UPM. I think it's better to deserialize; if we're concerned about compatibility, we can error on lockfile versions that are a major version higher than expected.

cdmistman (changes requested):
see feedback above

cdmistman:
closed in favor of #110

internal/backends/nodejs/nodejs.go
DillonB07: I did test using `pnpm init` and found that it doesn't have an option to generate with the defaults, but instead throws the user into a dialog, where they have to confirm options. Perhaps I made a typo and accidentally used `npm` when testing that, because that doesn't seem to happen if I test it again.

internal/backends/nodejs/nodejs.go
DillonB07: No, I didn't consider the other package manager commands as I tried to do everything in a similar way to how the system for yarn was implemented (hence the regex) and I'm also not very familiar with Go.

internal/backends/nodejs/nodejs.go
cdmistman: yeah, `pnpm` works for me without any interaction.

internal/backends/nodejs/nodejs.go
cdmistman: no worries! i am gonna shell out to pnpm after all - too much effort to deserialize lmao

PR #109, opened 2023-06-01
Use tree-sitter for JS parsing

cdmistman:
a user ran into a bug with UPM where it panicked due to being unable to parse the following js:
```js
const foo = ({
  // fails to parse the default argument here
  bar = () => {},
}) => {};
```

this is due to [amasad/esparse](https://github.com/amasad/esparse) being outdated. instead of relying on a stripped fork, which takes too much effort to maintain, it was decided we should use tree-sitter instead.

using tree-sitter is quite pleasant. despite the bugs referenced in the PR (smacker/go-tree-sitter#110 and smacker/go-tree-sitter#111), the experience was pleasant, and the performance is good (we do still have to cross an FFI boundary, though). we should probably switch to tree-sitter for all of the front-ends where we parse source files for guessing.


lincoln-replit (commented):
This seems alright to me. What can we do to be confident it won't cause regressions?

cdmistman:
> (I just realized I missed require('foo',) in the test, lemme fix that really quickly)

added it to the test but it didn't fail :)

lincoln-replit (approved):
Looks like we don't have tests set up in semaphore so I'll take your word for it. Do we have any indicators in prod for when panics or other problems happen?

cdmistman:
ah, yeah we should run go tests in CI. I'll make a linear

> Do we have any indicators in prod for when panics or other problems happen?
Not as far as I know. That's something that's better tackled in the workspace though. I'll make a Linear for this as well

masad-frost:
@replit/micromanager repl

PR #110, opened 2023-06-02
PNPM support

cdmistman:
thanks @DillonB07 for getting this started!

closes #108

### DIFF
--- a/internal/backends/nodejs/nodejs.go
+++ b/internal/backends/nodejs/nodejs.go
@@ -351,18 +351,35 @@ var NodejsPNPMBackend = api.LanguageBackend{
 	},
 	ListSpecfile: nodejsListSpecfile,
 	ListLockfile: func() map[api.PkgName]api.PkgVersion {
-		contentsB, err := ioutil.ReadFile("pnpm-lock.yaml")
+		// pnpm does scoped dependencies, so only list the dependencies at the top level
+		workspaceSpecContents, err := exec.Command("pnpm", "list", "--json", "--depth", "0").Output()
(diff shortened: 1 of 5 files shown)


internal/backends/nodejs/nodejs.go
```diff
@@ -309,6 +309,84 @@ var NodejsYarnBackend = api.LanguageBackend{
...
+	Lock: func() {
+		util.RunCmd([]string{"pnpm", "install"})
+	},
+	Install: func() {
+		util.RunCmd([]string{"pnpm", "install"})
+	},
+	ListSpecfile: nodejsListSpecfile,
+	ListLockfile: func() map[api.PkgName]api.PkgVersion {
+		// pnpm does scoped dependencies, so only list the dependencies at the top level
+		workspaceSpecContents, err := exec.Command("pnpm", "list", "--json", "--depth", "0").Output()
```
masad-frost, line 355: I imagine we actually want `--depth Inifnity` since we probably want to see all packages. This currently mirrors the specfile, I think. I'm not sure tbh what's the expected return from lock file listing, but other places list all sub dependencies.

cdmistman: although this brings up a good edge condition for a future UPM improvement: UPM expects a flat list of dependencies here, but pnpm and cargo (2 off the top of my head) support hierarchical dep trees

cdmistman: updated comment to reflect this state

masad-frost (approved):

PR #111, opened 2023-06-14
always use a correct User-Agent

cdmistman:
We should be Good Internet Citizens who Set User-Agent Correctly. This makes it Really Easy to do so.

### DIFF
--- a/internal/backends/python/gen_pypi_map/package_index.go
+++ b/internal/backends/python/gen_pypi_map/package_index.go
@@ -2,8 +2,9 @@ package main
 
 import (
 	"bufio"
-	"net/http"
 	"regexp"
+
+	"github.com/replit/upm/internal/api"
 )
 
 type PackageIndex struct {
@@ -26,15 +27,8 @@ func FakePackageIndex(packages ...string) PackageIndex {
 }
 
 func NewPackageIndex(index string, limit int) (PackageIndex, error) {
-	req, err := http.NewRequest("GET", index, nil)
-	if req != nil {
-		return PackageIndex{}, err
-	}
-
-	req.Header.Add("User-Agent", "upm (+https://github.com/replit/upm)")
-	resp, err := http.DefaultClient.Do(req)
-
-	if err != nil {
+	resp, err := api.HttpClient.Get(index)
+	if resp == nil || err != nil {
 		return PackageIndex{}, err
 	}
 
--- a/internal/backends/python/gen_pypi_map/package_index.go
+++ b/internal/backends/python/gen_pypi_map/package_index.go
@@ -26,7 +26,13 @@ func FakePackageIndex(packages ...string) PackageIndex {
 }
 
 func NewPackageIndex(index string, limit int) (PackageIndex, error) {
-	resp, err := http.Get(index)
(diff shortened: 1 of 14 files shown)


internal/backends/python/gen_pypi_map/package_index.go
```diff
@@ -26,7 +26,13 @@ func FakePackageIndex(packages ...string) PackageIndex {
...
 
 func NewPackageIndex(index string, limit int) (PackageIndex, error) {
-	resp, err := http.Get(index)
+	req, err := http.NewRequest("GET", index, nil)
+	if req != nil {
+		return PackageIndex{}, err
+	}
+
+	req.Header.Add("User-Agent", "upm (+https://github.com/replit/upm)")
+	resp, err := http.DefaultClient.Do(req)
```
ryantm, line 35:
```suggestion
	resp, err := api.HttpClient.Get(index)
```
Any reason we can't do this?

internal/backends/python/gen_pypi_map/test_modules.go
```diff
@@ -105,7 +105,12 @@ func TestModules(packages PackageIndex, bigqueryFile string, cacheDir string, pk
 }
 
 func GetPackageMetadata(packageName string) (PackageData, error) {
-	resp, err := http.Get("https://pypi.org/pypi/" + packageName + "/json")
+	req, err := http.NewRequest("GET", "https://pypi.org/pypi/" + packageName + "/json", nil)
+	if err != nil {
+		return PackageData{}, err
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
ryantm, line 113:
```suggestion
	resp, err := api.HttpClient.Get("https://pypi.org/pypi/" + packageName + "/json")
```

ryantm (approved):

internal/backends/python/gen_pypi_map/package_index.go
cdmistman: i don't believe so! i'll try it

internal/backends/python/gen_pypi_map/package_index.go
```diff
@@ -26,9 +27,8 @@ func FakePackageIndex(packages ...string) PackageIndex {
 }
 
 func NewPackageIndex(index string, limit int) (PackageIndex, error) {
-	resp, err := http.Get(index)
-
-	if err != nil {
+	resp, err := api.HttpClient.Get(index)
+	if resp == nil || err != nil {
```
ryantm, line 31: Why do we have to check for a nil resp here but not in most of the other places we use Get?

cdmistman: good question. fixing

cdmistman: i think the docs are indicating that `resp` can't be `nil` if `err` is non-`nil`. I'll just do a check in `HttpClient.Do` to assert that that's true, and if it isn't then panic (since without the check it would probably be a segfault due to accessing `nil` anyways)

ryantm (approved):

cdmistman:
manually confirmed each backend works identically to the version on `main`

PR #114, opened 2023-07-17
fix require import detection

cdmistman:
# why
1. When there isn't a string input to a require statement we are guessing an invalid package
2. It doesn't install any of the packages if one fails

# what changed
changed the tree-sitter query for javascript to only accept `require` imports if they're a single string

image

### DIFF
--- a/internal/backends/nodejs/grab.go
+++ b/internal/backends/nodejs/grab.go
@@ -72,7 +72,7 @@ func parseFile(contents []byte, results chan parseResult) {
 (
 	(call_expression
 		function: (identifier) @function
-		arguments: (arguments ((_) @other-imports)? ((_) @import) .))
+		arguments: (arguments ((_) @other-imports)? ((string) @import) .))
 	(#eq? @function "require"))
 `
 


masad-frost (approved):
Can we throw in a test for this please? Let's get a few degenerate cases

cdmistman:
> Can we throw in a test for this please?

can i defer this in a linear ticket? right now there's 0 behavioral testing infra for upm and i don't want that to block this pr

masad-frost (commented):
Do we not also guess dynamic imports? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import

cdmistman:
> do we not also guess dynamic imports?

ah, i guess not. i can handle that in another PR, but this is more urgent

masad-frost:
Oh what, that's news to me. Yeah let's track testing as a ticket.

masad-frost:
I'll defer review to Ryan

ryantm (approved):
Your testing screenshot looked good to me.

PR #116, opened 2023-07-24
Externalize pypi map into a ro sqlite database

airportyh:
## Why?

Having to load in the large maps within pypi_map.gen.go file into memory at start is causing upm 0.2 - 0.3 seconds.
Context: https://github.com/replit/goval/pull/9988#issuecomment-1646262367
We can eliminate this slowdown by externalizing this data into a lightweight database.

## Changes

1. Replaced gen_code.go with gen_db.go which generates a read-only sqlite database containing the equivalent data
2. created pypi_map.go - convenience wrapper for accessing the db's data
3. updated all accesses to any map in pypi_map.gen.go to accessing the same data using pypi_map.go instead
4. updated upm.nix to generate the db and setup an env var to point to its location for pypi_map.go to access it

Operations to be re-tested:
* python search
* python package guess

Performance enhancement should affect all languages and all operations.

## Testing

1. `nix build`
2. `mkdir play; cd play`
3. create `main.py` containing:
```python
import yaml
import requests
import flask
```
4. `time ../result/bin/upm lock`
6. `time ../result/bin/upm guess -f`

If you compare to the version in head, should see a sizable decrease in run time. In my local testing I saw ~200x for lock (when no op) and ~60x for guess.

Also, re-test `upm guess -f -l python` and `upm search -l python <query>` to make sure those still work.


ryantm (commented):
```
$ tree result
result
├── bin
│   ├── gen_pypi_map
│   └── upm
└── pypi_map.sqlite
```
Do we need the bin/gen_pypi_map file? It is pretty big.

airportyh:
> p file? It is pretty

That's only used for dev. I can look into getting rid of it. Probably a config to the buildGoModule call.

internal/backends/python/gen_pypi_map/db_gen.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Write all data within one transaction for speed
+	// https://stackoverflow.com/questions/1711631/improve-insert-per-second-performance-of-sqlite
+	_, err = db.Exec(`begin transaction;`)
+	if err != nil {
+		return err
+	}
+
+	// Guess at every module, add the guess and the package that was guessed to
+	// the masp
```
ryantm, line 82:
```suggestion
	// the map
```

internal/backends/python/pypi_map.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func NewPypiMap() (*PypiMap, error) {
+	dbFilePath := os.Getenv("PYPI_MAP_DB")
+	if dbFilePath == "" {
+		return nil, fmt.Errorf("PYPI_MAP_DB not set. Set it to the path of the sqlite database.")
+	}
+	// Open db in read-only mode so it doesn't try to make any modifications
+	// which would cause it to copy on write in our layered fs
+	db, err := sql.Open("sqlite3", dbFilePath+"?mode=ro")
```
ryantm, line 28: Immutable should be faster and safe here because we are storing this int he Nix store, which won't change out from under us. https://www.sqlite.org/uri.html#recognized_query_parameters

```suggestion
	db, err := sql.Open("sqlite3", dbFilePath+"?immutable=1")
```

[Cribbed from the Layered store work ](https://github.com/NixOS/nix/pull/8356/files#diff-766e649c52f1b2f334ad2788aa2230b88b0a7cdd2799529c6125462960b145e0R508)

ryantm: We're definitely going to want to do this. I just tried it and it is faster (TM).

ryantm (changes requested):

internal/backends/python/pypi_map.go
ryantm: Before on my computer:
```
[nix-shell:~/p/replit/upm/play]$ time ../result/bin/upm lock

real  0m0.081s
user  0m0.086s
sys 0m0.037s

[nix-shell:~/p/replit/upm/play]$ time ../result/bin/upm guess -fa
--> python3 /run/user/1000/upm3791543655/bare-imports.py '<secret sauce>'
flask
pyyaml
requests

real  0m0.124s
user  0m0.108s
sys 0m0.044s
```
After:

```
[nix-shell:~/p/replit/upm/play]$ time ../result/bin/upm lock

real  0m0.014s
user  0m0.006s
sys 0m0.009s

[nix-shell:~/p/replit/upm/play]$ time ../result/bin/upm guess -fa
--> python3 /run/user/1000/upm2231193298/bare-imports.py '<secret sauce>'
flask
pyyaml
requests

real  0m0.041s
user  0m0.030s
sys 0m0.013s
```

With immutable:

```
[nix-shell:~/p/replit/upm/play]$ time ../result/bin/upm lock

real  0m0.006s
user  0m0.002s
sys 0m0.005s

[nix-shell:~/p/replit/upm/play]$ time ../result/bin/upm guess -fa
--> python3 /run/user/1000/upm2868831452/bare-imports.py '<secret sauce>'
flask
pyyaml
requests

real  0m0.028s
user  0m0.023s
sys 0m0.007s
```

ryantm (approved):

PR #118, opened 2023-08-15
search mjs and cjs files for js guessing

cdmistman:
https://linear.app/replit/issue/DX-167/mjs-files-are-not-guessed


ryantm (approved):
Thanks!

PR #119, opened 2023-08-18
Implemented pkg search with pypi; porting asadmoosvi/pypi-search

airportyh:
# Why?

We are going around upm currently for Python package search and instead using https://pipsearch.replit.app/ which is simply an in-house wrapper around https://github.com/asadmoosvi/pypi-search/blob/main/pypi_search/search.py. Simpler would be just reimplement pypi_search in upm.

Context: https://replit.slack.com/archives/C03KS2B221W/p1692041472039169

## Change

Ported pypi_search/search.py

## Test

1. `make upm`
2. `./cmd/upm/upm -l python search requests`


internal/backends/python/search_pypi.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/replit/upm/internal/api"
+	"golang.org/x/net/html"
+)
+
+var WhiteSpaceChars = " \n\t"
+
+// Port of https://github.com/asadmoosvi/pypi-search/blob/main/pypi_search/search.py
+func SearchPypi(query string) ([]api.PkgInfo, error) {
+	endpoint := fmt.Sprintf("http://pypi.org/search/?q=%s", url.QueryEscape(query))
```
ryantm, line 17:
```suggestion
	endpoint := fmt.Sprintf("https://pypi.org/search/?q=%s", url.QueryEscape(query))
```

I already tested this works.

internal/backends/python/search_pypi.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/replit/upm/internal/api"
+	"golang.org/x/net/html"
+)
+
+var WhiteSpaceChars = " \n\t"
+
+// Port of https://github.com/asadmoosvi/pypi-search/blob/main/pypi_search/search.py
+func SearchPypi(query string) ([]api.PkgInfo, error) {
+	endpoint := fmt.Sprintf("https://pypi.org/search/?q=%s", url.QueryEscape(query))
+	req, err := http.NewRequest("GET", endpoint, nil)
```
ryantm, line 18: Can you use your http client definitions in api/http.go instead? They will use our user-agent.

ryantm (changes requested):

ryantm (commented):
Tried it out on my computer looks good.

@replit/micromanager imagegen circumventing the busted server

ryantm (approved):

PR #123, opened 2023-09-08
update things

cdmistman:
it's been a while since things have been updated here, which means it's chore time!

- use go 1.20
- update nixpkgs input


ryantm (approved):
tried to build locally and it worked, changes look good too, thanks!

PR #127, opened 2023-09-26
nix: standardize flake, add devshell, format

ryantm:
Why
===
* We want to have a nix devshell
* We don't need to use flake-utils
* Let's have a Nix formatter

What changed
===
* Remove flake-utils
* Remove default.nix
* Adopt a somewhat more standard directory structure for nix files
* Add a go devshell
* Use Alejandra for nix fmt

Test plan
===
* upm builds with `nix build -L` and has a reasonable version string


masad-frost (approved):
Noice

PR #131, opened 2023-10-03
pnpm tests

cdmistman:
there's a bug with reading the lockfile (since we directly serialize it), which is why tests are failing


internal/backends/nodejs/nodejs.go
```diff
@@ -357,36 +358,35 @@ var NodejsPNPMBackend = api.LanguageBackend{
...
-				}
-
-				depVersion, ok := depInfo.(map[string]interface{})["version"].(string)
-				if !ok {
-					util.Die("pnpm list: %s", err)
-				}
-
-				pkgs[pkgName] = api.PkgVersion(depVersion)
+		switch lvMajor {
+		case "6":
```
cdmistman, line 377: decided to only focus on supporting 6.x since repl-it-web started with 6.x lockfile

internal/backends/nodejs/nodejs.go
```diff
@@ -357,36 +358,35 @@ var NodejsPNPMBackend = api.LanguageBackend{
...
+			dependencies, ok := lockfile["dependencies"]
+			if !ok {
+				return pkgs
+			}
+
+			for pkgName, pkgInfo := range dependencies.(map[interface{}]interface{}) {
+				// i love golang
```
ryantm, line 384:
```suggestion
```

ryantm (approved):

PR #126, opened 2023-09-25
test suite

cdmistman:
cucumber was too much - felt like i was debugging the library more than upm itself

this supercedes #124

PR #132, opened 2023-10-03
npm tests

cdmistman:
we're not reading packages from the lockfile at all, which is why tests are failing

blocked on #126


internal/backends/nodejs/nodejs.go
cdmistman, line 1: this is a bugfix

test-suite/templates/nodejs-npm/many-deps/package-lock.json
```diff
@@ -0,0 +1,2241 @@
...
+      "license": "MIT",
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    }
+  }
+}
```
ryantm, line 2241: sheesh.

ryantm (approved):
Man those lock files for tests are big. I tried making some invalid package-lock.json files and it didn't blow up.

cdmistman:
yeah, the test framework presumes that the lockfiles are valid, since they're an implementation detail of the package manager itself. however, i'm wary of manually editing them - things should Just Work. the only reason they're committed is to reduce lockfile resolution time. it's faster if the pm doesn't have to do any resolution

PR #133, opened 2023-10-06
remove statik

cdmistman:
# why

statik sucks. keeps breaking things and there's an ***embed***ded tool for this in go itself (see what i did there?)

# what changed

- changed use of statik to `go:embed`
- as a result, created `github.com/replit/upm/resources` package
  - NB it's public facing


airportyh (approved):
Nice!

PR #135, opened 2023-10-12
add install-replit-nix-system-dependencies subcommand

ryantm:
Why
===
* When installing Python packages, we want to add system dependencies to replit.nix, so the packages work.
* This can eventually be expanded to other languages too.

What changed
===
* Introduce InstallReplitNixSystemDependencies subcommand. The subcommand uses the language specfile and packages sent on the commandline (the new ones we're adding) to determine what it should try to look up in the system dependency map.
* Make a default function that does nothing for all languages except Python
* For Python, add a python_map.json that maps package names to system dependency install operations and embed in go
* Add tests for functions that translate the map into nix-editor ops

Test plan
===
* Remove replit.nix so we have a blank slate to start
* Make a pyproject.toml file like
```
[tool.poetry.dependencies]
python = ">=3.10.0,<3.11"
psycopg2 = "^1.26.0"
```
* run `REPL_HOME=. go run ./cmd/upm install-replit-nix-system-dependencies`
* Confirm the replit.nix file has a postgres dep in the deps and and python ld library path.
* run `REPL_HOME=. go run ./cmd/upm install-replit-nix-system-dependencies pycairo`
* Confirm that cairo and pkg-config are added to deps

### DIFF
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -68,3 +68,23 @@ func ReplitNixAddToNixEditorCmds(replitNixAdd ReplitNixAdd) [][]string {
 
 	return result
 }
+
+func RunNixEditorCmds(cmds [][]string) {
+	for _, cmd := range cmds {
+		output := util.GetCmdOutput(cmd)
+
+		var nixEditorStatus struct {
+			Status string
+			Data   string
+		}
+
+		if err := json.Unmarshal(output, &nixEditorStatus); err != nil {
+			util.Die("unexpected nix-editor output: %s", err)
+		}
+		if nixEditorStatus.Status != "success" {
+			util.Die("nix-editor error: %s", nixEditorStatus.Data)
+		}
+		// otherwise we have success and don't need to output
+		// anything
+	}
+}
--- a/internal/nix/nix.go
+++ b/internal/nix/nix.go
@@ -1,0 +1,70 @@
+package nix
+
+import (
+	_ "embed"
+	"encoding/json"
+	"log"
+	"os"
+	"path"
+
+	"github.com/replit/upm/internal/api"
+	"github.com/replit/upm/internal/util"
+)
+
+type ReplitNixAdd struct {
+	Deps              []string `json:"deps,omitempty"`
+	PythonLibraryDeps []string `json:"libdeps,omitempty"`
+}
+
+var (
+	python_map_var map[string]ReplitNixAdd
+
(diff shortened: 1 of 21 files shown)


internal/nix/nix.go
```diff
@@ -0,0 +1,90 @@
...
+
+	for _, dep := range replitNixAdd.PythonLibraryDeps {
+		result = append(result, []string{"nix-editor", "--path", path, "--dep-type", "python", "--add", dep})
+	}
+
+	return result
+}
+
+func RunNixEditorCmds(cmds [][]string) {
+	for _, cmd := range cmds {
```
cdmistman, line 73: nix-editor supports commands over stdin right? can we make this a single shell-out and pass the commands over stdin instead?

ryantm: Yes, but I view that nix-editor interface to be overkill and sort of want to kill it. If we find performance issues with this, happy to do it.

cdmistman: i think that interface is perfect in this scenario. `ReplitNixAddToNixEditorCmds` has to be careful to avoid a race condition (if you executed nix-editor in that function in a goroutine you'd have a race-condition). having an IPC interface is perfect for repetitive execution of very-small actions, and we'll be able to add optimizations later wrt OT

cdmistman (approved):

internal/nix/nix.go
```diff
@@ -0,0 +1,142 @@
...
+		err := encoder.Encode(op)
+		if err != nil {
+			util.Die("unable to turn op into json: %v error: %s", op, err)
+		}
+	}
+	_, err = stdin.Write(in.Bytes())
+	if err != nil {
+		util.Die("unable to write to nix-editor")
+	}
+	stdin.Close()
```
cbrewster, line 115:
```suggestion
	encoder := json.NewEncoder(stdin)
	for _, op := range ops {
		err := encoder.Encode(op)
		if err != nil {
			util.Die("unable to turn op into json: %v error: %s", op, err)
		}
	}
	err = stdin.Close()
	if err != nil {
		util.Die("unable to write to nix-editor")
	}
```
No need for the temporary buffer

cdmistman (approved):
nice, do you mind writing a test :P

ryantm:
> nice, do you mind writing a test :P

Yeah, we have a unit test here, but I'll add an integration test in a follow on.

PR #137, opened 2023-10-13
Remove x/crypto package, use x/term

masad-frost:
Supercedes #97 

# What changed
- Switch to using term package instead of terminal inside crypto
- Remove crypto package since it's not used anywhere

# Test plan
- `upm search -l nodejs-npm --format table leftpad`
- works and has the right terminal width

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,8 +12,8 @@ require (
 	github.com/smacker/go-tree-sitter v0.0.0-20230501083651-a7d92773b3aa
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/testify v1.8.1
-	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
 	golang.org/x/net v0.5.0
+	golang.org/x/term v0.4.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -12,8 +12,8 @@ require (
 	github.com/smacker/go-tree-sitter v0.0.0-20230501083651-a7d92773b3aa
 	github.com/spf13/cobra v0.0.5
 	github.com/stretchr/testify v1.8.1
-	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
-	golang.org/x/net v0.5.0
+	golang.org/x/crypto v0.14.0
```
ryantm, line 15: I think we can remove x/crypto after this. It only was used in table.go.

ryantm (approved):

go.mod
masad-frost: Oh, for some reason I thought crypto/md5 comes from that, but it's part of standard lib. Will fix, nice catch.

PR #141, opened 2023-10-20
python_map: add mysqlclient, pyzbar sort

ryantm:
Why
===
* We need more system dependency map entries to make it useful.

What changed
===
* Sorted the list
* Added pyzbar
* Added mysqlclient

Test plan
===
* confirm Nix unit tests still pass: go test -v -race ./internal/nix/...


cdmistman (approved):
lgtm

PR #145, opened 2023-10-24
more guess tests

cdmistman:
#109 fixed a bug, i added a regression test for it

i also forgot to include the guess test files for #144, that's been addressed here

### DIFF
--- a/test-suite/templates/guess/js/esparse-fail
+++ b/test-suite/templates/guess/js/esparse-fail
@@ -1,0 +1,7 @@
+import express from 'express';
+
+const foo = ({
+	// esparse fails to parse the default argument here,
+	// resulting in the full guess failing
+	bar = () => { },
+}) => { };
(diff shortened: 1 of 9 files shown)


test-suite/templates/guess/js/esparse-fail
ryantm, line 1: I'm confused by the name of this file. Is the guess supposed to fail?

masad-frost: It would fail pre #109

ryantm (approved):

test-suite/templates/guess/js/esparse-fail
cdmistman: no, it's a regression test. esparse failed to parse the file. since there's no issue to link to, do you have a better name? https://martinfowler.com/bliki/TwoHardThings.html

cdmistman: oops, i didn't even see faris's response when i sent this. apologies!

PR #146, opened 2023-10-25
add explicit require test to js guess tests

cdmistman:
just realized the only test using `require` is in dedup


ryantm (approved):
Thanks!

PR #147, opened 2023-10-25
axe python2 backend and break out a couple functions, fix tests

cdmistman:
the python impl is a little hard to read/scroll through so i broke a couple functions out of the make* function

i also axed the python2 backend since python2 is dead, long live python3!

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -95,68 +95,99 @@ func normalizePackageName(name api.PkgName) api.PkgName {
 	return api.PkgName(nameStr)
 }
 
-// pythonMakeBackend returns a language backend for a given version of
+func info(name api.PkgName) api.PkgInfo {
+	res, err := api.HttpClient.Get(fmt.Sprintf("https://pypi.org/pypi/%s/json", string(name)))
+
+	if err != nil {
+		util.Die("HTTP Request failed with error: %s", err)
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode == 404 {
+		return api.PkgInfo{}
+	}
+
+	if res.StatusCode != 200 {
+		util.Die("Received status code: %d", res.StatusCode)
+	}
+
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		util.Die("Res body read failed with error: %s", err)
+	}
+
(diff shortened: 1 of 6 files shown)


ryantm (approved):

cdmistman:
oh, *OH* the tests from #144 haven't run until now because i was switching on the wrong name in the test suite 🤦 the test-suite checks for the backend called `python3-poetry` which was actually `python-python3-poetry` until this PR

internal/backends/python/python.go
```diff
@@ -186,10 +214,15 @@ func pythonMakeBackend(name string, python string) api.LanguageBackend {
 			// be a pretty easy fix, though. (Why is this
 			// so complicated??)
 
-			outputB := util.GetCmdOutput([]string{
+			outputB, err := util.GetCmdOutputFallible([]string{
 				"poetry",
 				"config", "settings.virtualenvs.path",
 			})
+			if err != nil {
+				// there's no virtualenv configured, so no package directory
```
ryantm, line 222: Do we need to detect .pythonlibs here somehow? We weren't before and it was working.

internal/util/command.go
```diff
@@ -33,14 +33,22 @@ func RunCmd(cmd []string) {
...
+// GetCmdOutputFallible prints and runs the given command, returning its
+// stdout as a string. Stderr goes to the terminal. GetCmdOutputFallible
+// does not exit the process on error or command failure, but instead
+// returns an error.
+func GetCmdOutputFallible(cmd []string) ([]byte, error) {
+	ProgressMsg(quoteCmd(cmd))
+	command := exec.Command(cmd[0], cmd[1:]...)
+	command.Stderr = os.Stderr
+	return command.Output()
+}
```
ryantm, line 45: nice extraction

ryantm (approved):

internal/backends/python/python.go
cdmistman: this is mostly to satisfy the test suite. i think something else is going on on replit, but `settings.virtualenvs.path` is broken with `.pythonlibs`
image

i'll take the conversation about how to resolve this for nixmodules offline. i don't think packager3 ever actually calls this function, so i'm not too worried about it being at least slightly broken for the time being

PR #148, opened 2023-10-25
[python nixpkgs map] add 700+ entries

ryantm:
Why
===
* If we're going to fix Python system dependencies, we need more entries.
* These entries were farmed from Nixpkgs, so they should be somewhat reliable.

Note about the methodology here: I extracted the function arguments to all the Python3 modules in Nixpkgs and filtered the build-support and darwin inputs, and a few others that I suspected were just build tools as opposed to runtime dependencies. I assembled them into a map and then tested that they all can build in a mkShell on nixpkgs 22.11 and 23.05 channels. This does not mean all of these will work. I added everything to libdeps that had a lib directory in its build output.

What changed
===
* Added tons more entries
* Checked that all the deps are cached in the nixpkgs 22.11 and 23.05 Replit channels

Test plan
===
* Make sure the pycairo test case still works
* Make sure Running a Typescript Repl still works


masad-frost (approved):
Yo this is awesome. I wonder how we can validate those.

ryantm:
> I wonder how we can validate those.

One way these are sort of validated is the pycairo, pyzbar, and psycopg2 entries are supersets of what they were.

PR #150, opened 2023-10-26
add Alias field to backends, support at CLI and store levels

cdmistman:
# Why

#147 made a couple of breaking changes to the CLI:
- removed `python-python2-poetry` backend
- simplified `python-python3-poetry` backend name to `python3-poetry`

unfortunately, the latter change will cause all upm caches to be invalidated, and invocations of `upm -l python-python3-poetry` will break.

# What changed

- Added `Alias` field to `api.LanguageBackend`
- When resolving the backend via `-l` flag, if resolving via `Name` of the `LanguageBackend` fails, try to resolve against the `Alias`
- The logic for reading from the cache store is now:
  - check if there's a match for `Name`, if there is then use that cache
  - check if there's a match for `Alias`, if there is then use that cache
- The logic for creating a cache entry in the store is now:
  - check if there's a match for `Name`, exit function if true
  - check if there's a match for `Alias`, exit function if true
  - create an entry for `Name`

### DIFF
--- a/internal/backends/backends.go
+++ b/internal/backends/backends.go
@@ -54,6 +54,7 @@ func matchesLanguage(b api.LanguageBackend, language string) bool {
 	for _, lPart := range strings.Split(language, "-") {
 		if !bParts[lPart] {
 			checkAlias = true
+			break
 		}
 	}
 	if checkAlias {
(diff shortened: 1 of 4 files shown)


internal/backends/backends.go
```diff
@@ -50,9 +50,21 @@ func matchesLanguage(b api.LanguageBackend, language string) bool {
 	for _, bPart := range strings.Split(b.Name, "-") {
 		bParts[bPart] = true
 	}
+	checkAlias := false
 	for _, lPart := range strings.Split(language, "-") {
 		if !bParts[lPart] {
-			return false
+			checkAlias = true
```
ryantm, line 56:
```suggestion
			checkAlias = true
			break
```

ryantm (approved):

PR #151, opened 2023-10-27
ignore all packages in /test-suite for packagers we test

cdmistman:
it's annoying getting dependabot PRs because we don't even actively use these packages. this will help minimize disk usage (by keeping the packages Eternally Pinned, we don't download ♾️ versions of packages just to run some tests), reduce unnecessary maintenance burden, and enhance predictability in our test suite.

note that dependabot doesn't have a way to blanket-ignore a directory. super annoying. tracked by https://github.com/dependabot/dependabot-core/issues/4364. this means we'll have to remember to go to https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem every time we test another backend, and remember to update the dependabot.yml file.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  # https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem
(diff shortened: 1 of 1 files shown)


airportyh (approved):
Thanks! til.

PR #152, opened 2023-10-27
improve js guessing

cdmistman:
- the query now handles dynamic imports
- the query now properly rejects `require` calls that don't have exactly 1 argument
- when getting the results of the query, always *just* get the `@import` tag


internal/backends/nodejs/grab.go
```diff
@@ -105,24 +106,18 @@ func parseFile(contents []byte, results chan parseResult) {
...
+			if query.CaptureNameForId(capture.Index) == "import" {
+				importPath = capture.Node.Content(contents)
+				break
 			}
+		}
 
-			importPath = match.Captures[1].Node.Content(contents)
+		if importPath == "" {
+			// shouldn't happen, with the way the query is written and handled.
+			util.Die("Failed to parse import path")
```
masad-frost, line 118: I feel like we don't have enough observability to react to these effectively, but maybe we fixed that

test-suite/templates/guess/js/import
masad-frost, line 1: Good to call this dynamici-mport

masad-frost (approved):
Nice

internal/backends/nodejs/grab.go
cdmistman: @ryantm has a WIP branch adding o11y to upm

PR #153, opened 2023-10-27
python guessing with treesitter

cdmistman:
# why

right now python guessing requires
1. making a temporary directory
2. writing a couple python scripts to said directory
3. synchronously walking the filesystem

this is quite expensive! OTOH, tree-sitter is performant and works in-process, so we get to 🪓 some performance pain

note: python guessing tests are relatively lackluster compared to js/ts guessing, i hope things don't break 😬 i don't know python very well and have done my best to test extensively, but i didn't even catch JS/TS dynamic imports breaking until ryan caught it (fixed in #152)

# what changed

- tree-sitter abstraction now supports reading pragmas


internal/backends/python/grab.go
```diff
@@ -0,0 +1,333 @@
...
+
+	foundImportPaths := map[string]bool{}
+	for _, pkg := range pkgs {
+		foundImportPaths[pkg] = true
+	}
+
+	return foundImportPaths, nil
+}
+
+func filterImports(foundPkgs map[string]bool) (map[api.PkgName]bool, bool) {
```
airportyh, line 274: Could you add a test for filtering imports as well?

cdmistman: that doesn't work with the unit tests, since the `PYPI_MAP_DB` env var needs to be set. the filtering functionality is tested with the test-suite, though

airportyh (approved):

PR #154, opened 2023-10-30
better treesitter ergonomics

cdmistman:
while moving python over to treesitter, i noticed a lot of the logic is the same. as such i decided to abstract this out into its own utility function

- created new `util.GuessWithTreeSitter` function which grabs all `@import` tags from  a tree-sitter query, language, and a directory path
- the above function supports using globs to find/filter paths to guess, which is a planned feature
- added another js test that tests builtin/runtime module imports


internal/backends/nodejs/grab.go
```diff
@@ -57,193 +54,122 @@ var internalModules = []string{
...
-			importPath = match.Captures[1].Node.Content(contents)
+		if pkgs[api.PkgName(mod)] {
+			continue
 		}
 
-		// TODO: https://github.com/smacker/go-tree-sitter/issues/111
-		// once the above issue is resolved, use the string destructuring used in the
-		// `import_statement` pattern above in the `call_expression` pattern instead
-		// of using this hacky trim
-		importPaths = append(importPaths, strings.Trim(importPath, "'\"`"))
```
cdmistman, line 126: still haven't ported over the quote stripping, and for some reason the `@org/package` import isn't working, but most of the output is correct :)

cdmistman:
image

WHY DIDN'T YOU TELL ME THAT BEFORE DEPENDABOT????!!!!?!?!?!?!?!?!!!!??

internal/backends/nodejs/grab.go
```diff
@@ -77,168 +75,104 @@ func parseFile(contents []byte, results chan parseResult) {
...
 	if err != nil {
-		results <- parseResult{nil, false}
-		return
+		util.Die("couldn't guess imports: %s", err)
 	}
 
-	qc.Exec(query, node)
+	modules := append([]string{}, jsPkgs...)
+	modules = append(modules, tsPkgs...)
+	modules = append(modules, tsxPkgs...)
```
ryantm, line 98: Is it useful to enforce uniqueness in the modules list here to avoid doing extra work later on?

internal/backends/nodejs/grab_test.go
```diff
@@ -1,60 +0,0 @@
...
-	}
-
-	for i, path := range result.importPaths {
-		if i >= len(expected) {
-			t.Errorf("Unexpected import %s", path)
-		} else if path != expected[i] {
-			t.Errorf("Expected %s, got %s", expected[i], path)
-		}
-	}
-}
```
ryantm, line 60: Why are we deleting this test? It's hard for me to see that the integration tests added cover these cases, and isn't it nice to have some unit tests for this feature?

cdmistman: the unit tests haven't been running in CI anyways. you're right though, i'll add the missing cases here

cdmistman: i'll note that the test is also wrong

```
require(` + "`child_process`" + `)
```

cdmistman: forgot to attach

cdmistman: it was a golang syntax highlighting issue, the above was valid. fixed!

internal/backends/nodejs/grab.go
cdmistman: done in [99bb975](https://github.com/replit/upm/pull/154/commits/99bb975f8a394b465a74e040c635d1c219128abb)

test-suite/templates/guess/js/original.expect
```diff
@@ -0,0 +1 @@
+nan
```
ryantm, line 1: Before we were expecting all of these but now just nan?

		"fs",
		"assert",
		"path",
		"nan",
		"buffer",
		"console",
		"dns",
		"child_process",

ryantm: I'm not comfortable removing the unit test. I feel like there is a bunch of parser functionality it is testing that I'm not confident is captured by the integration tests. Can we not remove it? Okay to not make it run for now, we can do that in a separate PR.

cdmistman: restored and passing in [2c33a83](https://github.com/replit/upm/pull/154/commits/2c33a83c8ab221c17fec19ab6f16b5bc6e256970)

ryantm (approved):

PR #155, opened 2023-11-01
speed-up builds with buildGoCache

Mic92:
While this takes of them time from the go build, it seems that the majority of time is still consumed by building the sqlite database.
I tried to isolate this code into its own derivation but the generator seems to have cross dependencies to the rest of the code base.


masad-frost:
@cdmistman rerouting this to you

nix/upm/imported-packages
```diff
@@ -0,0 +1,46 @@
+archive/tar
```
Mic92, line 1: My recommendation for this would be to have a github action that just updates this file after pushes to master automatically. After all it's a performance optimization and it causes too much friction/merge conflicts to have this being updated in pull requests.

Mic92: Something like that but for updating this file: https://gist.github.com/Broxzier/1ed980b8b3822d213e98042fc6a92040

cdmistman (approved):
2.5 mins down to 20s - amazing! thank you!

cdmistman:
@replit/micromanager merge CI isn't running because semaphore

PR #162, opened 2023-11-03
fix dependabot

cdmistman:
it's annoying and apparently the config i set up before was full of errors. this time i had my yaml extension working so it validated against the schema, and i double-checked the docs. hopefully dependabot schema checker for CI kicks in on this PR as affirmation that we won't get stupid dependabot PRs anymore


ryantm (approved):
:facepalm: dependabot

PR #165, opened 2023-11-14
Datadog msgs log to /tmp/upm.log

airportyh:
# Why?

Don't want to [display DD logs](https://github.com/replit/goval/pull/11405) to the console.

## Changes

This writes them in `/tmp/upm.dd.log`

### DIFF
--- a/internal/trace/logger.go
+++ b/internal/trace/logger.go
@@ -1,0 +1,27 @@
+package trace
+
+import "os"
+
+type DatadogLogger struct {
+	file *os.File
+}
+
+func NewDatadogLogger() (*DatadogLogger, error) {
+	file, err := os.Create("/tmp/upm.dd.log")
+	if err != nil {
+		return nil, err
+	}
+
+	return &DatadogLogger{
+		file: file,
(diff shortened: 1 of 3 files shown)


internal/trace/logger.go
```diff
@@ -0,0 +1,27 @@
+package trace
+
+import "os"
+
+type DatadogLogger struct {
+	file *os.File
+}
+
+func NewDatadogLogger() (*DatadogLogger, error) {
+	file, err := os.Create("/tmp/upm.dd.log")
```
ryantm, line 10: Are we okay with only having the logs from the latest upm run. Probably fine, but we could parameterize by process id or something to make it more unique.

ryantm (approved):

internal/trace/logger.go
airportyh: I didn't want to potentially create a lot of files and make it harder to figure out which one is the latest. Any other down side to this?

ryantm: Nah. Fine to keep it like this.

PR #167, opened 2023-11-16
[python map] add reflex

ryantm:
Why
===
* We found that reflex init needs unzip to work

What changed
===
* Added unzip

Test plan
===
* go test ./internal/nix/... still works
* on replit, try installing reflex and observe unzip in replit.nix


airportyh (approved):
:)

PR #168, opened 2023-11-28
check regex match results when looking for a pragma

cdmistman:
# why

upm was panicking due to index-out-of-range when checking for upm pragmas. turns out the code as-is expects there to _always_ be a valid pragma if there's just a comment associated with the import line

# what changed

when reading captures for the upm pragma, check that there's a regexp match in the first place

# test plan

tests pass


ryantm (approved):
Thanks for the quick fix! Could we add a test to exercise this code path?

cdmistman:
already did, was running locally to confirm it passed :)

PR #169, opened 2023-12-05
Search result heuristics

blast-hardcheese:
Attempt to surface more relevant packages to the top of the search results.

### DIFF
--- a/internal/api/types.go
+++ b/internal/api/types.go
@@ -203,6 +203,10 @@ type LanguageBackend struct {
 	// which packages are installed. The path need not exist.
 	GetPackageDir func() string
 
+	// Apply a sensible heuristic for sorting search results
+	// if we know we want to surface some packages over others.
+	SortPackages func(query string, ignoredPackages []string, packages []PkgInfo) []PkgInfo
+
 	// Search for packages using an online index. The query may
 	// contain any characters, including whitespace. Return a list
 	// of search results, which can be of any length. (It will be
--- a/internal/api/types.go
+++ b/internal/api/types.go
@@ -3,12 +3,21 @@ package api
 import (
 	"context"
 	"regexp"
+	"strings"
 
 	"github.com/replit/upm/internal/util"
 )
 
 // PkgName represents the name of a package, e.g. "flask" for Python.
 type PkgName string
+
+func (n PkgName) Contains(other PkgName) bool {
+	return strings.Contains(string(n), string(other))
+}
+
+func (n PkgName) HasPrefix(other PkgName) bool {
+	return strings.HasPrefix(string(n), string(other))
+}
 
 // PkgSpec represents a package version constraint, e.g. "^1.1" or ">=
 // 1.1, <2.0" for Python.
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/BurntSushi/toml"
@@ -16,6 +17,8 @@ import (
 	"github.com/replit/upm/internal/util"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
+
+var normalizationPattern = regexp.MustCompile(`[-_.]+`)
 
 // this generates a mapping of pypi packages <-> modules
 // moduleToPypiPackage pypiPackageToModules are provided
@@ -84,10 +87,11 @@ func normalizeSpec(spec interface{}) string {
 
 // normalizePackageName implements NormalizePackageName for the Python
 // backends.
+// See https://packaging.python.org/en/latest/specifications/name-normalization/
 func normalizePackageName(name api.PkgName) api.PkgName {
 	nameStr := string(name)
 	nameStr = strings.ToLower(nameStr)
-	nameStr = strings.Replace(nameStr, "_", "-", -1)
+	nameStr = normalizationPattern.ReplaceAllString(nameStr, "-")
 	return api.PkgName(nameStr)
 }
 
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -14,6 +14,7 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/replit/upm/internal/api"
 	"github.com/replit/upm/internal/nix"
+	"github.com/replit/upm/internal/pkg"
 	"github.com/replit/upm/internal/util"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
@@ -256,6 +257,10 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
 
 			return filepath.Join(path, base+"-py"+version)
 		},
+		SortPackages: func(query string, ignoredPackages []string, packages []api.PkgInfo) []api.PkgInfo {
+			return pkg.SortPrefixSuffix(normalizePackageName, query, ignoredPackages, packages)
+		},
+
 		Search: func(query string) []api.PkgInfo {
 			results, err := SearchPypi(query)
 			if err != nil {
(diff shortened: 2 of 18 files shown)


internal/pkg/pkg.go
```diff
@@ -0,0 +1,53 @@
...
+
+	ignoredPackageSet := makeLoweredHM(b, ignoredPackages)
+
+	needle := b.NormalizePackageName(api.PkgName(query))
+	var exact *api.PkgInfo
+	prefixed := []api.PkgInfo{}
+	infixed := []api.PkgInfo{}
+	filtered := []api.PkgInfo{}
+
+	// If we have an exact package match, ensure it comes first.
```
blast-hardcheese, line 28: Should probably change this comment, now that we're shuffling results quite a bit more.

internal/api/types.go
```diff
@@ -3,13 +3,22 @@ package api
 import (
 	"context"
 	"regexp"
+	"strings"
 
 	"github.com/replit/upm/internal/util"
 )
 
 // PkgName represents the name of a package, e.g. "flask" for Python.
 type PkgName string
```
airportyh, line 12: Why declare this type instead of simply using string?

internal/backends/python/python.go
```diff
@@ -84,10 +85,15 @@ func normalizeSpec(spec interface{}) string {
 
 // normalizePackageName implements NormalizePackageName for the Python
 // backends.
+// See https://packaging.python.org/en/latest/specifications/name-normalization/
 func normalizePackageName(name api.PkgName) api.PkgName {
 	nameStr := string(name)
 	nameStr = strings.ToLower(nameStr)
-	nameStr = strings.Replace(nameStr, "_", "-", -1)
+	matcher, err := regexp.Compile(`[-_.]+`)
```
airportyh, line 92: We can use regexp.MustCompile and set a global variable at the module level to avoid re-compiling.

internal/pkg/pkg.go
```diff
@@ -0,0 +1,53 @@
...
+			prefixed = append(prefixed, pkg)
+		} else if lower.Contains(needle) {
+			infixed = append(infixed, pkg)
+		} else {
+			filtered = append(filtered, pkg)
+		}
+	}
+
+	curated := []api.PkgInfo{}
+	if exact != nil {
```
airportyh, line 45: How good is this scheme compared to the default? Have we analyze the quality of the results for some common queries? Are there situations where this scheme could demote some more desirable results?

internal/pkg/pkg.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, pkg := range names {
+		normal := b.NormalizePackageName(api.PkgName(pkg))
+		set[normal] = struct{}{}
+	}
+	return set
+}
+
+func SortPackages(b api.LanguageBackend, query string, ignoredPackages []string, packages []api.PkgInfo) []api.PkgInfo {
+
+	ignoredPackageSet := makeLoweredHM(b, ignoredPackages)
```
airportyh, line 20: I don't think this initialization loop is necessary. You just want a map of bool, I think.

internal/api/types.go
blast-hardcheese: Opaque types are a good way to communicate intent through types as well as avoid bugs during refactors.

As an example, `map[PkgName]PkgInfo` means that all the symantics we want in a `PkgName` have been satisfied, presumably normalization being the most important.

That would keep `BeautifulSoup4` and `beautifulsoup4` from being duplicates in that map.

internal/backends/python/python.go
blast-hardcheese: Ah, good one! I'll make the change

internal/pkg/pkg.go
blast-hardcheese: I think it depends on the backend, pypi surfaces particularly low quality results due to "relevance" being heavily weighted by the release date.

Maybe other package searches are better at this, to be sure, though in my experience `${ecosystem}-${extension}` is a common way to associate packages, with `${extension}-${ecosystem }` being less common but not impossible (thinking of a bridge between two different ecosystems, as an example). This is why I have both of them called out explicitly as higher ranking results.

By accumulating these sorted relevant packages and maintaining sub-ordering, my hope is that we have something of the best of both worlds, with exact name matches being on the top for relevance.

internal/pkg/pkg.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"github.com/replit/upm/internal/api"
+)
+
+func makeLoweredHM(b api.LanguageBackend, names []string) map[api.PkgName]struct{} {
+	// Build a hashset. struct{}{} purportedly is of size 0, so this is as good as we get
+	set := make(map[api.PkgName]struct{})
+	for _, pkg := range names {
+		normal := b.NormalizePackageName(api.PkgName(pkg))
```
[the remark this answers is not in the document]
blast-hardcheese: In my research, a map of empty structures is preferred to a map of `bool` because bool takes memory where the empty structure consumes zero bytes. Inclusion in the hashset is sufficient, I don't need an additional boolean

internal/pkg/pkg.go
blast-hardcheese: Testing out `--language npm react`, most of the packages are still in the same results list, but stuff like `rxjs` is below the rest of the standard `react` libraries. A downside is that we don't search descriptions, so `styled-components` and `prop-types` are outside of the grouping.

Before
---
image

After
---
image

airportyh: Since pypi is the one singled out as problematic, I'd prefer to only apply this to pypi.

airportyh: If it makes sense, I'd move the sorting logic into python's search function.

internal/pkg/pkg.go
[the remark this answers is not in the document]
blast-hardcheese: Resolved over slack! I didn't know about default values in Go, relying on that default value makes this read much cleaner.

internal/api/types.go
airportyh: 👍

internal/pkg/pkg.go
```diff
@@ -0,0 +1,56 @@
...
+	set := make(map[api.PkgName]bool)
+	for _, pkg := range names {
+		normal := normalizePackageName(api.PkgName(pkg))
+		set[normal] = true
+	}
+	return set
+}
+
+func SortPrefixSuffix(normalizePackageName func(api.PkgName) api.PkgName, query string, ignoredPackages []string, packages []api.PkgInfo) []api.PkgInfo {
+	ignoredPackageSet := makeLoweredHM(normalizePackageName, ignoredPackages)
```
airportyh, line 23: I thought we didn't need this anymore.
```suggestion
	ignoredPackageSet := make(map[string]bool)
```

internal/pkg/pkg.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"github.com/replit/upm/internal/api"
+)
+
+func SortNoop(query string, ignoredPackages []string, packages []api.PkgInfo) []api.PkgInfo {
+	return packages
+}
+
+func makeLoweredHM(normalizePackageName func(api.PkgName) api.PkgName, names []string) map[api.PkgName]bool {
+	// Build a hashset. struct{}{} purportedly is of size 0, so this is as good as we get
```
airportyh, line 13:
```suggestion
	// Build a hashset. struct{}{} purportedly is of size 0, so this is as good as we get
```
```suggestion
```

internal/pkg/pkg.go
```diff
@@ -0,0 +1,56 @@
...
+	set := make(map[api.PkgName]bool)
+	for _, pkg := range names {
+		normal := normalizePackageName(api.PkgName(pkg))
+		set[normal] = true
```
airportyh, line 17: I actually didn't realize you were setting this to true...

airportyh (approved):
Thank you!

internal/pkg/pkg.go
blast-hardcheese: @airportyh Still need to run `normalizePackageName` on the packages, I can in-line though if you'd prefer

blast-hardcheese: I see the approval -- Thank you!

PR #170, opened 2023-12-05
feat/Ignore packages in SortNoop

blast-hardcheese:
This got overlooked in review comments from #169 

- Delete `SortNoop`
- Promote `--ignored-packages` to top-level, so all backends get it for free

### DIFF
--- a/internal/pkg/pkg.go
+++ b/internal/pkg/pkg.go
@@ -5,20 +5,6 @@ import (
 	"github.com/replit/upm/internal/api"
 )
 
-func NormalizePackageNameNoop(name api.PkgName) api.PkgName {
-	return name
-}
-
-func SortNoop(normalizePackageName func(api.PkgName) api.PkgName) func(query string, packages []api.PkgInfo) []api.PkgInfo {
-	return func(query string, packages []api.PkgInfo) []api.PkgInfo {
-		return sortNoop(normalizePackageName, query, packages)
-	}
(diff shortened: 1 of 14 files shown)


airportyh (commented):
Didn't occur to me earlier but seems weird to have the SortPackages function also be responsible for filtering packages that were previously installed. Could we move the filtering out side? I think it might simplify things. Also seems wrong that the exact matching is [done before the filtering](https://github.com/replit/upm/blob/main/internal/pkg/pkg.go#L35).

internal/pkg/pkg.go
```diff
@@ -5,23 +5,27 @@ import (
...
+	return name
 }
 
-func makeLoweredHM(normalizePackageName func(api.PkgName) api.PkgName, names []string) map[api.PkgName]bool {
-	// Build a hashset. struct{}{} purportedly is of size 0, so this is as good as we get
-	set := make(map[api.PkgName]bool)
-	for _, pkg := range names {
-		normal := normalizePackageName(api.PkgName(pkg))
-		set[normal] = true
+func SortNoop(normalizePackageName func(api.PkgName) api.PkgName) func(query string, packages []api.PkgInfo) []api.PkgInfo {
```
airportyh, line 12: and this higher-order function pattern is probably unneeded.

airportyh (approved):
:+1:

PR #176, opened 2023-12-11
more accurately deserialize `package.json` files

cdmistman:
we're getting upm panics because upm can't handle the (very common) `"author": "string"` form in `package.json` files, instead expecting it to be `"author": { "name": "string", "email": "string" }`. i've noticed this happening for a few other package.json fields so these should also be fixed


internal/backends/nodejs/nodejs.go
```diff
@@ -48,24 +50,121 @@ type npmSearchResults struct {
 //
 // See https://github.com/npm/registry/blob/5db1bb329f554454467531a3e1bae5e97da160df/docs/responses/package-metadata.md
 // for documentation on the format.
+//
+// ---
+//
+// note by @cdmistman: the format isn't 100% validated to the type that's described in the link above.
```
cdmistman, line 56:
```suggestion
// note: the format isn't 100% validated to the type that's described in the link above.
```

internal/backends/nodejs/nodejs.go
```diff
@@ -48,24 +50,121 @@ type npmSearchResults struct {
...
+	URL   string
+}
+
+var packageJsonPersonStringRegexp *regexp.Regexp
+
+func (person *packageJsonPerson) UnmarshalJSON(data []byte) error {
+	var str string
+	if err := json.Unmarshal(data, &str); err == nil {
+		if packageJsonPersonStringRegexp == nil {
+			packageJsonPersonStringRegexp = regexp.MustCompile(`([^\<\(]*)?\s*(\<.*\>)?\s*(\(.*\))?`)
```
ryantm, line 107: Instead of relying on a regex here, could we Unmarshall into a interface{} type and process that? Not sure if it is better, but I think worth exploring if you haven't already.

internal/backends/nodejs/nodejs.go
```diff
@@ -48,24 +50,124 @@ type npmSearchResults struct {
...
+		}
+
+		if url, ok := obj["url"].(string); ok {
+			person.URL = url
+		}
+
+		return nil
+	}
+
+	return errors.New("expected person in package.json to be a string or an object")
```
ryantm, line 148: If we fail to parse the author, can we continue with an empty author instead of failing? I'm worried about the string regex failing to parse some author strings and then we still fail.

ryantm (changes requested):

internal/backends/nodejs/nodejs.go
cdmistman: i can add a fallback in the path where the regexp fails to parse! i don't feel comfortable changing this return though, as it's the case where the author field is neither a string nor an object, which _should_ be an error. handling deserialization errors is the responsibility further up the stack from this function, we shouldn't expect this function to always succeed

ryantm: Sounds good.

ryantm (approved):
Yay, less package search pain, thank you!

PR #178, opened 2023-12-11
[ci] fix lint

ryantm:
Why
===
* CI lint was broken in https://replit.semaphoreci.com/jobs/e535bc22-e0fe-4f4f-bd11-97b4e856a801
* Upgrading golangci-lint might fix it, since it looks like the problems were for the wrong version of Go

What changed
===
* Bump version go golangci-lint we install in CI

Test plan
===
* This CI run


cdmistman (approved):
nice, thanks

PR #179, opened 2023-12-12
[poetry] Make show-package-dir work again

blast-hardcheese:
Remove manual reimplementation of poetry's venv location logic in favor of just emitting nothing if the directory doesn't exist.

Previous logic had drifted away from how poetry actually works. If the previous spec'd behavior is desired, we can open a change against poetry itself to surface this information, or attempt to use poetry as a library to get access to internals.

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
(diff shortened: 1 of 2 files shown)


airportyh:
@blast-hardcheese could you add testing instructions?

blast-hardcheese:
Closing in favor of #182, I didn't realize there were two.

PR #180, opened 2023-12-13
pip support

blast-hardcheese:
Depends on #183, [Minified diff](https://github.com/replit/upm/compare/dstewart/permit-nonreproducible-backends-to-be-tested...dstewart/pip-support)

Draft, pending making constraints work, as well as tests

What's been manually tested:

- `upm install`
- `upm add <pkg> [pkg...]`
- `upm list`

seems to work well as a first pass.

For notes about the pip lockfile, please refer to [this discussion](https://discuss.python.org/t/what-is-the-status-on-a-pip-lock-file/6669/8)


internal/backends/python/python.go
```diff
@@ -347,7 +333,176 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+		},
+		ListSpecfile: func() map[api.PkgName]api.PkgSpec {
+			flags, result := ListRequirementsTxt("requirements.txt")
+
+			// Stash the seen flags into a module global
+			pipFlags = flags
+
+			return result
+		},
+		GuessRegexps: util.Regexps([]string{
```
masad-frost, line 472: Should this be shared with the poetry backend?

internal/backends/python/python.go
```diff
@@ -347,7 +333,176 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+		})
+	}
+	pipPath := path.Join(venvPath, "bin", "pip")
+	return venvPath, pipPath
+}
+
+// makePythonPipBackend returns a backend for invoking poetry, given an arg0 for invoking Python
+// (either a full path or just a name like "python3") to use when invoking Python.
+func makePythonPipBackend(python string) api.LanguageBackend {
+	var pipFlags []PipFlag
```
masad-frost, line 350: Not sure what assumptions we're making here, but the way UPM is used is by executing a command and exiting, we won't have the luxury of using something stored in memory

internal/backends/python/requirements.go
```diff
@@ -0,0 +1,167 @@
...
+		} else if name, spec, found := findPackage(line); found {
+			// Found a package!
+			sofar[*name] = *spec
+		} else if nextfile, found := strings.CutPrefix(line, "-r "); found {
+			// # It is possible to refer to other requirement files...
+			// -r other-requirements.txt
+			var newFlags []PipFlag
+			newFlags, sofar, constraints = recurseRequirementsTxt(depth+1, strings.TrimSpace(nextfile), sofar, constraints)
+			flags = append(flags, newFlags...)
+		} else if _, found := strings.CutPrefix(line, "-c "); found {
```
masad-frost, line 105: TIL about constraints.

masad-frost: Oh man, reading the requirements.txt spec, yikes.

masad-frost (commented):
Noice progress

internal/backends/python/python.go
```diff
@@ -347,7 +333,176 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+			span, ctx := tracer.StartSpanFromContext(ctx, "pip install")
+			defer span.Finish()
+
+			// Create venv if it does not exsit
+			venv := guessVirtualenvPath()
+			if venv == "" {
+				venv = fallbackVenvDirectory
+			}
+			var pipPath string
+			venv, pipPath = ensureVenvExists(python, venv)
```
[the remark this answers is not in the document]
airportyh: We don't use venv mostly because there are hardcoded paths to python inside it, preventing us from patch upgrading python in the future.

internal/backends/python/python.go
blast-hardcheese: It was either this or add an additional `interface {}` parameter to every invocation of the `Add` function, since the pip backend needs information only derived during `ListSpecfile`. Colton and I kicked back and forth on a few options before we settled on this one, it works because `ListSpecfile` is always run before `Add`.

Perhaps I can also add a `len(...) > 0` check as well, to catch the case where it hadn't been run yet, or we just parse the specfile twice in a row and are OK with that, if you've got a preference I'm happy to hear it -- my preference would be to just add the `interface {}`, but that's a change with a high blast radius.

internal/backends/python/python.go
```diff
@@ -347,7 +333,176 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+// (either a full path or just a name like "python3") to use when invoking Python.
+func makePythonPipBackend(python string) api.LanguageBackend {
+	var pipFlags []PipFlag
+
+	b := api.LanguageBackend{
+		Name:                 "python3-pip",
+		Specfile:             "requirements.txt",
+		Alias:                "python-python3-pip",
+		FilenamePatterns:     []string{"*.py"},
+		Quirks:               api.QuirksAddRemoveAlsoInstalls | api.QuirksCannotLock | api.QuirksNotReproducible,
```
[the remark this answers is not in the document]
blast-hardcheese: Yeah, I could definitely see that. The reason I didn't go that way was because it would be defining yet another bespoke standard. If we're guaranteed to be running inside Conda, perhaps relying on Conda to do locking makes sense, but then that just adds more interdependency spaghetti.

internal/backends/python/python.go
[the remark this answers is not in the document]
blast-hardcheese: `site.USER_SITE` it is.

internal/backends/python/python.go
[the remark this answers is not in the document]
blast-hardcheese: Resolved over Slack. Lockfiles should provide guarantees that are outside of the scope of what `requirements.txt` offers. Perhaps https://discuss.python.org/t/what-is-the-status-on-a-pip-lock-file/6669/8 will resolve this someday.

test-suite/Install_test.go
```diff
@@ -8,11 +8,20 @@ import (
...
+	"python3-poetry": true,
+	"python3-pip":    true,
+}
+
 func TestInstall(t *testing.T) {
 	for _, bt := range languageBackends {
 		bt.Start(t)
 
-		if bt.Backend.Name != "bun" && bt.Backend.Name != "nodejs-npm" && bt.Backend.Name != "nodejs-pnpm" && bt.Backend.Name != "nodejs-yarn" && bt.Backend.Name != "python3-poetry" {
+		if testInstalls[bt.Backend.Name] {
```
cdmistman, line 24:
```suggestion
		if !testInstalls[bt.Backend.Name] {
```

internal/backends/python/python.go
```diff
@@ -311,6 +304,167 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+				cmd = append(cmd, string(name))
+			}
+			util.RunCmd(cmd)
+			err := RemoveFromRequirementsTxt("requirements.txt", pkgs)
+			if err != nil {
+				util.Die("%s", err.Error())
+			}
+		},
+		Install: func(ctx context.Context) {
+			//nolint:ineffassign,wastedassign,staticcheck
```
cdmistman, line 420: what's this?

internal/backends/python/requirements.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		if line == "" {
+			// Skip blank lines
+		} else if name, spec, found := findPackage(line); found {
+			// Found a package!
+			sofar[*name] = *spec
+		} else if nextfile, found := strings.CutPrefix(line, "-r "); found {
+			// # It is possible to refer to other requirement files...
+			// -r other-requirements.txt
```
cdmistman, line 103: 🤯

internal/backends/python/requirements.go
cdmistman: python always makes me feel like i'm tripping

internal/backends/python/requirements.go
```diff
@@ -0,0 +1,192 @@
...
+			// ... or constraints files.
+			// -c constraints.txt
+			//
+			// TODO: Accumulate constraints to pass to underlyling pip
+		} else if parts := strings.SplitN(line, " ", 2); len(parts) > 1 && knownFlags[parts[0]] {
+			flags = append(flags, PipFlag(strings.Join(parts, " ")))
+			// If we find an editable package, try to extract the package name out of the URI
+			// This both shows that package in the `upm list` output, as well as prevents upm
+			// from inserting it into requirements.txt, which would then cause a conflict
+			// on the next run.
```
cdmistman, line 120: nice

Makefile
```diff
@@ -91,6 +91,9 @@ endif
 ifdef UPM_CI
 test-suite:
 	go get gotest.tools/gotestsum
+	venv="$$(mktemp -d)"; \
+	python -m venv "$$venv"; \
+	source "$$venv/bin/activate"; \
```
cdmistman, line 96: is there a different way to do this? i'd rather be able to support multiple venvs being used in a test suite run

internal/cli/cmds.go
```diff
@@ -362,10 +362,10 @@ func runRemove(language string, args []string, upgrade bool,
 
 	// Map from normalized package names to original package
 	// names.
-	normPkgs := map[api.PkgName]api.PkgName{}
```
cdmistman, line 365: any reason to not keep this as api.PkgName?

blast-hardcheese: I'm starting a slog towards `api.PkgName` being exclusively used for normalized package names. It bit me a few times in this PR, expecting that values would match when one or the other wasn't normalized. Open for other strategies here, but `string` vs `api.PkgName` seemed like the most sensible option without lots of churn.

Makefile
blast-hardcheese: Yeah -- this was what motivated my comment over Slack about before/after's -- I don't like this either, especially since it pollutes the environment between test cases. I just wanted to see ✅ at least once before delving back into the ❌'s

internal/backends/python/python.go
```diff
@@ -311,6 +304,167 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+
+				cmd = append(cmd, name+spec)
+			}
+			// Run install
+			util.RunCmd(cmd)
+			// Determine what was actually installed
+			{
+				outputB, err := util.GetCmdOutputFallible([]string{
+					"pip", "freeze",
+				})
```
cdmistman, line 375: here's one of my gripes about upm not having defined semantics. technically we're locking packages here! which means pip is, in fact, reproducible! it just so happens that the spec file(s) can also be lock file(s)... maybe we shouldn't be marking pip as non-reproducible? and maybe for the `Lock` func we can run `pip freeze` and replace all the lines without a constraint with the constrained version...

test-suite/utils/upm.go
```diff
@@ -8,9 +8,14 @@ import (
 )
 
 func (bt *BackendT) UpmAdd(pkgs ...string) {
-	beforeLockDeps := bt.UpmListLockFile()
+
```
cdmistman, line 11:
```suggestion
```

internal/cli/cmds.go
cdmistman: ah, makes sense :) i miss rust...

test-suite/utils/upm.go
```diff
@@ -138,6 +149,17 @@ func (bt *BackendT) UpmInstall() {
...
+	for idx, pkg := range pkgs {
+		out := (string)(bt.Backend.NormalizePackageName(api.PkgName(pkg.Name)))
+		if pkgs[idx].Name != out {
+			bt.t.Log("Inconsistently normalized package name in results: ", pkgs[idx].Name, "did not equal", out)
+		}
+		pkgs[idx].Name = out
+	}
+	return pkgs
+}
+
```
cdmistman, line 162: let's address the normalization stuff in another PR? i wanna have a better view of how much logic this affects and where we have the wrong logic for normalized package names

test-suite/templates/python3-pip/many-deps/requirements.txt
cdmistman, line 1: can we mix up the constraints a bit so we can get more common-case test coverage?

cdmistman: fine to do this in another PR

test-suite/utils/upm.go
blast-hardcheese: Ope, merged in #183. So, the reason I added this here is to ensure the tests don't break, a subsequent PR should actually _remove_ this, once we're normalized.

Makefile
cdmistman: gotcha. can you add a TODO to move our command invocation to being inside of a shell, where we can define before/afters?

internal/backends/python/python.go
cdmistman: from the lens of a behavioral descriptivist (given we haven't prescribed *any* behavior for upm at all, lmao), i still think we should have `Lock` for the pip backend. upm's behavior for `upm lock` has been to allow the package manager to ensure that all dependencies are locked. if dependencies aren't locked, then they should be

internal/backends/python/python.go
```diff
@@ -311,6 +304,167 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+			if venv := os.Getenv("VIRTUAL_ENV"); venv != "" {
+				return venv
+			}
+
+			if outputB, err := util.GetCmdOutputFallible([]string{
+				"python",
+				"-c", "import site; print(site.USER_SITE)",
+			}); err == nil {
+				return string(outputB)
+			}
```
cdmistman, line 340: should we check if a venv is available before falling back to user site?

internal/backends/python/python.go
```diff
@@ -311,6 +304,167 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+		},
+		SortPackages: pkg.SortPrefixSuffix(normalizePackageName),
+
+		Search: func(query string) []api.PkgInfo {
+			results, err := SearchPypi(query)
+			if err != nil {
+				util.Die("failed to search pypi: %s", err.Error())
+			}
+			return results
+		},
```
cdmistman, line 352: can probably break this out into `func search` to keep things DRY and make it more clear that this behavior is identical to poetry backend's

internal/backends/python/python.go
blast-hardcheese: I'm happy to go along with this, save for where to put the secondary `requirements.txt` (and how to tie it in).

We have some choices:
1. One, singular `requirements.txt` that has all the locked transitive deps. Now the user has transitive deps that need to be solved when either bumping or uninstalling libraries.
2. One `requirements.txt` that has the `upm add`'d deps, another `requirements.lock.txt` (or `.full.txt` or whatever), which gets used by `upm install`, but that breaks user expectations if switching between `upm` and `pip` (Do I `pip install -r requirements.txt` or `pip install -r requirements.lock.txt`?)
3. `requirements.txt` _is_ the `pip freeze` file, and `upm add` adds to/removes from `requirements.upm.txt` or similar

Given the ambiguity and hassle for the user, I would actually prefer to leave locking in the hands of the user until https://discuss.python.org/t/what-is-the-status-on-a-pip-lock-file/6669/8 .

Benefits to deferring:
- Those that know/care about reproducibility can choose to lock in their deployments.
- Those that are using docker to build anyway get dependency locking inherent to their production images.
- Removing the burden of maintaining the transitive deps keeps transitive deps unpinned, guiding users towards a glorious future of maintainability for their projects instead of "Nothing works! I need to delete my requirements.txt" (which was very close to what we saw with `replit-py` the other week)

internal/backends/python/python.go
```diff
@@ -311,6 +304,167 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+				result[normalizePackageName(name)] = spec
+			}
+
+			// Stash the seen flags into a module global
+			pipFlags = flags
+
+			return result
+		},
+		GuessRegexps: pythonGuessRegexps,
+		Guess:        func(ctx context.Context) (map[api.PkgName]bool, bool) { return guess(ctx, getPython3()) },
```
cdmistman, line 443:
```suggestion
		Guess:        func(ctx context.Context) (map[api.PkgName]bool, bool) { return guess(ctx, python) },
```

internal/backends/python/python.go
blast-hardcheese: I had that initially, but removed it in favor of https://github.com/replit/upm/pull/180#discussion_r1425890670

I agree with this, actually, since if the user `activate`'s the venv before running `upm add`, they'll install into the venv, if they don't they won't. Compare that to only _one_ subcommand activating the venv, giving inconsistent behavior.

internal/backends/python/python.go
blast-hardcheese: good catch. I thought I got them all.

internal/backends/python/python.go
cdmistman: good point!

internal/backends/python/python.go
cdmistman: gotcha. then we probably shouldn't lock the deps we add at all right? since that's more likely to cause breakages in the package?

internal/backends/python/python.go
```diff
@@ -311,6 +306,161 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+			util.RunCmd([]string{"pip", "install", "-r", "requirements.txt"})
+		},
+		ListSpecfile: func() map[api.PkgName]api.PkgSpec {
+			flags, _result, err := ListRequirementsTxt("requirements.txt")
+			if err != nil {
+				util.Die("%s", err.Error())
+			}
+
+			result := make(map[api.PkgName]api.PkgSpec)
+			for name, spec := range _result {
```
cdmistman, line 429:
```suggestion
			flags, result, err := ListRequirementsTxt("requirements.txt")
			if err != nil {
				util.Die("%s", err.Error())
			}

			result := make(map[api.PkgName]api.PkgSpec)
			for name, spec := range result {
```

internal/backends/python/requirements.go
```diff
@@ -0,0 +1,194 @@
...
+				err := recurseRemoveFromRequirementsTxt(depth+1, nextfile, pkgs)
+				if err != nil {
+					return err
+				}
+				lines = append(lines, line)
+			} else {
+				lines = append(lines, line)
+			}
+		}
+	}
```
cdmistman, line 170: kinda weird to do anonymous blocks in go
```suggestion
	for scanner := bufio.NewScanner(handle); scanner.Scan() {
		line := strings.TrimSpace(scanner.Text())

		if name, _, found := findPackage(line); found && pkgs[normalizePackageName(*name)] {
			continue
		} else if nextfile, found := strings.CutPrefix(line, "-r "); found {
			err := recurseRemoveFromRequirementsTxt(depth+1, nextfile, pkgs)
			if err != nil {
				return err
			}
			lines = append(lines, line)
		} else {
			lines = append(lines, line)
		}
	}
```

internal/backends/python/requirements.go
```diff
@@ -136,6 +136,17 @@ func ListRequirementsTxt(path string) ([]PipFlag, map[api.PkgName]api.PkgSpec, e
...
+func cutPrefixes(line string, options ...string) (nextfile string, found bool) {
+	for _, choice := range options {
+		nextfile, found = strings.CutPrefix(line, choice)
+		if found {
+			break
+		}
+	}
+
+	return nextfile, found
+}
```
cdmistman, line 148: can probably go in util tbh

internal/backends/python/python.go
```diff
@@ -311,6 +306,161 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
+			for name, spec := range pkgs {
+				name := string(name)
+				spec := string(spec)
+
+				cmd = append(cmd, name+spec)
+			}
+			// Run install
+			util.RunCmd(cmd)
+			// Determine what was actually installed
+			{
```
cdmistman, line 368: can you get rid of this anonymous block as well? it's the end of the function anyways, not hiding any vars from unrelated logic

cdmistman (approved):
besides last 2 nits lgtm!

internal/backends/python/python.go
blast-hardcheese: I think the particulars of the package are fine, but transitive deps are not OK, same as in many other languages. If something 5 layers down has a CVE, we should permit that update to make its way through naturally.

The benefit is giving some stability in which library API we're targeting, so if we `upm add` right before a major version bump our code doesn't start breaking immediately after we create the project.

PR #182, opened 2023-12-15
Fix poetry show package dir

blast-hardcheese:
How poetry calculates the env path has changed, so instead of trying to guess, just delegate to `poetry env`. If the environment hasn't been created we don't get a path, which is only relevant if we're in a completely blank environment, so it's hopefully not that bad.

- Renaming Python3Backend to `PythonPoetryBackend`
- Rely on `poetry env list` to find the active environment
- General housekeeping

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -203,25 +202,8 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
 				return venv
 			}
 
-			// Ideally Poetry would provide some way of
-			// actually checking where the virtualenv will
-			// go. But it doesn't. So we have to
-			// reimplement the logic ourselves, which is
-			// totally fragile and disgusting. (No, we
-			// can't use 'poetry run which python' because
-			// that will *create* a virtualenv if one
-			// doesn't exist, and there's no workaround
-			// for that without mutating the global config
-			// file.)
-			//
-			// Note, we don't yet support Poetry's
-			// settings.virtualenvs.in-project. That would
-			// be a pretty easy fix, though. (Why is this
-			// so complicated??)
-
 			outputB, err := util.GetCmdOutputFallible([]string{
-				"poetry",
-				"config", "virtualenvs.path",
+				"poetry", "env", "list", "--full-path",
 			})
 			if err != nil {
 				// there's no virtualenv configured, so no package directory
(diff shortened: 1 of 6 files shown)


blast-hardcheese:
A question from @airportyh around how to test this seems to have been eaten by GitHub, but the following is an example:

```
upm (nix) ((f547cb7...)) dstewart$ cd /tmp/poetry-test
poetry-test (nix) dstewart$ ~/Projects/upm/cmd/upm/upm show-package-dir
--> poetry env list --full-path
poetry-test (nix) dstewart$ ~/Projects/upm/cmd/upm/upm add requests
--> poetry add requests
Creating virtualenv poetry-test-cX6xfNfC-py3.10 in /Users/dstewart/Library/Caches/pypoetry/virtualenvs
Using version ^2.31.0 for requests

Updating dependencies
Resolving dependencies... (0.2s)

Package operations: 5 installs, 0 updates, 0 removals

  • Installing certifi (2023.11.17)
  • Installing charset-normalizer (3.3.2)
  • Installing idna (3.6)
  • Installing urllib3 (2.1.0)
  • Installing requests (2.31.0)

Writing lock file
poetry-test (nix) dstewart$ ~/Projects/upm/cmd/upm/upm show-package-dir
--> poetry env list --full-path
/Users/dstewart/Library/Caches/pypoetry/virtualenvs/poetry-test-cX6xfNfC-py3.10
```
Since we are delegating the `show-package-dir` subcommand directly to `poetry env list`, the virtualenv doesn't exist until the first time a package operation has been performed, hence the result being empty in the first invocation.

internal/backends/python/python.go
```diff
@@ -203,59 +202,24 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
-				base = strings.ToLower(filepath.Base(cwd))
-			}
-
-			version := strings.TrimSpace(string(util.GetCmdOutput([]string{
-				python, "-c",
-				`import sys; print(".".join(map(str, sys.version_info[:2])))`,
-			})))
-
-			return filepath.Join(path, base+"-py"+version)
+			return path
```
cdmistman, line 222: is this backwards-compatible? we need to be able to support several poetry versions - ideally *at least* the minimum of the current major version

blast-hardcheese: It is backwards compatible with the earliest 1.x version of poetry I can find on nixpkgs, 1.0.2. Unfortunately it is _not_ compatible with [0.12.16](https://github.com/python-poetry/poetry/releases/tag/0.12.16), though that's from May 2019, nearly 4 years old at this point. The reason I'm making this change is because the current functionality doesn't work at all for any 1.x version of poetry, to the best of my knowledge.

```
[nix-shell:/tmp/boop]# poetry add reqeuests
Creating virtualenv boop-s3yOBcnp-py3.7 in /root/.cache/pypoetry/virtualenvs

...

[nix-shell:/tmp/boop]# poetry env list --full-path
/root/.cache/pypoetry/virtualenvs/boop-s3yOBcnp-py3.7 (Activated)

[nix-shell:/tmp/boop]# poetry --version
Poetry version 1.0.2
```

cdmistman (approved):
cool! thanks!

PR #181, opened 2023-12-15
Permit individual suites

blast-hardcheese:
Permit running individual language suites via `make test-...`

Example:

```bash
$ make test-python3  # Run all python3 backends
...
$ make test-python3-poetry  # Run only the python3-poetry backend
...
$ make test-nodejs  # Run all nodejs backends
Preparing test suites:
- nodejs-npm
- nodejs-pnpm
- nodejs-yarn
```

PR #183, opened 2023-12-16
Permit nonreproducible backends to be tested

blast-hardcheese:
~Depends on #181, #182 [Minified diff](https://github.com/replit/upm/compare/5b1ff50...dstewart/permit-nonreproducible-backends-to-be-tested)~

- Relax some restrictions around how we do testing when the nonreproducible quirk is enabled on a backend

### DIFF
--- a/test-suite/Remove_test.go
+++ b/test-suite/Remove_test.go
@@ -49,21 +49,39 @@ func doRemove(bt testUtils.BackendT, templatesToPkgsToRemove map[string][]string
 
 		if tmpl != "no-deps" {
 			bt.Subtest(tmpl, func(bt testUtils.BackendT) {
-				bt.Subtest("locked", func(bt testUtils.BackendT) {
-					bt.Subtest("each", func(bt testUtils.BackendT) {
-						bt.AddTestFile(template+bt.Backend.Specfile, bt.Backend.Specfile)
-						bt.AddTestFile(template+bt.Backend.Lockfile, bt.Backend.Lockfile)
-						for _, pkg := range pkgsToRemove {
-							bt.UpmRemove(pkg)
-						}
-					})
+				if bt.Backend.QuirksIsReproducible() {
+					bt.Subtest("locked", func(bt testUtils.BackendT) {
+						bt.Subtest("each", func(bt testUtils.BackendT) {
+							bt.AddTestFile(template+bt.Backend.Specfile, bt.Backend.Specfile)
+							bt.AddTestFile(template+bt.Backend.Lockfile, bt.Backend.Lockfile)
+							for _, pkg := range pkgsToRemove {
+								bt.UpmRemove(pkg)
+							}
+						})
 
-					bt.Subtest("all", func(bt testUtils.BackendT) {
-						bt.AddTestFile(template+bt.Backend.Specfile, bt.Backend.Specfile)
-						bt.AddTestFile(template+bt.Backend.Lockfile, bt.Backend.Lockfile)
-						bt.UpmRemove(pkgsToRemove...)
+						bt.Subtest("all", func(bt testUtils.BackendT) {
+							bt.AddTestFile(template+bt.Backend.Specfile, bt.Backend.Specfile)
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -95,6 +95,32 @@ test-suite:
...
+#                                # since it doesn't know where upm or the PYPI_MAP_DB are.
+test-%:
+	nix develop -c nix shell -c make wrapped-$@
+
+unwrapped-test-%:
+	label="$@"; \
+	label="$${label#unwrapped-}"; \
+	PATH=$$PWD/cmd/upm/:"$$PATH" \
+	PYPI_MAP_DB=$$PWD/internal/backends/python/pypi_map.sqlite \
+			 make wrapped-$$label
```
cdmistman, line 117: this is different from #181, can you rebase or merge into that pr?

internal/backends/python/python.go
```diff
@@ -203,59 +202,24 @@ func makePythonPoetryBackend(python string) api.LanguageBackend {
...
-				base = strings.ToLower(filepath.Base(cwd))
-			}
-
-			version := strings.TrimSpace(string(util.GetCmdOutput([]string{
-				python, "-c",
-				`import sys; print(".".join(map(str, sys.version_info[:2])))`,
-			})))
-
-			return filepath.Join(path, base+"-py"+version)
+			return path
```
cdmistman, line 222: see my comment on #182

Makefile
blast-hardcheese: Oops, automerge burned me. I'll split it out.

internal/backends/python/python.go
blast-hardcheese: Replied there as well

test-suite/Remove_test.go
```diff
@@ -49,20 +49,22 @@ func doRemove(bt testUtils.BackendT, templatesToPkgsToRemove map[string][]string
 
 		if tmpl != "no-deps" {
 			bt.Subtest(tmpl, func(bt testUtils.BackendT) {
-				bt.Subtest("locked", func(bt testUtils.BackendT) {
```
cdmistman, line 52: oof, actually there should be a locked an unlocked variant. can you keep this locked subtest? we can add unlocked in another PR, but good to keep this as a TODO

test-suite/WhichLanguage_test.go
```diff
@@ -24,11 +24,18 @@ func TestWhichLanguage(t *testing.T) {
...
+					bt.AddTestFile(template+bt.Backend.Specfile, bt.Backend.Specfile)
+					bt.AddTestFile(template+bt.Backend.Lockfile, bt.Backend.Lockfile)
+					bt.UpmWhichLanguage()
+				})
+			} else {
+				bt.Subtest("no lockfile", func(bt testUtils.BackendT) {
+					bt.AddTestFile(template+bt.Backend.Specfile, bt.Backend.Specfile)
+					bt.UpmWhichLanguage()
+				})
+			}
```
cdmistman, line 38: i think we can just get rid of this `else`, there's nothing different vs the default case. can't exactly test multiple non-reproducible package managers that use the same spec file 😬

Makefile
blast-hardcheese: https://github.com/replit/upm/pull/187

test-suite/Remove_test.go
blast-hardcheese: I can fix that here if you'd prefer

test-suite/WhichLanguage_test.go
blast-hardcheese: Sorry, maybe missing something -- the `default` case only curently applies to `bun`, are you suggesting I add `python3-pip` to that list? "default" doesn't really mean anything useful in the pip case, do you have insight into what it could be generalized into?

test-suite/Remove_test.go
blast-hardcheese: Fixed

test-suite/WhichLanguage_test.go
cdmistman: ah, you're right, lgtm

cdmistman (approved):

PR #186, opened 2023-12-18
Adding release-drafter

blast-hardcheese:
Bringing visibility to upm version compatibility


ryantm (approved):
Let's try it. Maybe we can use an auto-updater internally.

PR #200, opened 2023-12-28
Define PkgSet abstraction

blast-hardcheese:
RFC on the direction of this changeset.

The motivation here is to insulate the different backends from incidental changes to the guess data model.

`map[string]bool` needs to become `map[string][]api.PkgName` in order to permit multiple packages to satisfy the same `import`, in languages that permit package collisions (Python, in this case. Maybe others later.)

Second PR incoming, for more context.


blast-hardcheese:
Closing this after reflowing #201 to not need `PkgSet`.

Issue #15, opened 2019-11-01
Imports incorrect modules in Python when Poetry already has correct module


AlexBlandin:
So some packages, such as `py-ulid`, have a different name from the imported module, in this case `import ulid`, or as usually done, `from ulid import ULID`.

If you import with Repl.it, then the first time you run it everything works fine. If you close and open again, the next time it tries to import the module `ulid`, which has a completely different interface and doesn't have a `ULID` class to import, and now no longer runs as `ulid`, which python attempts to resolve against rather than `py-ulid` (and doesn't keep checking after the failure), doesn't have a `ULID` to import.

Namely, UPM needs to either check against the existing packages that have been added, seeing if they are applicable, or it needs to check whether the packages it tried to include actually satisfy the interface of the program (even just flagging up that the `ulid` package doesn't have `ULID` while `py-ulid`, which is already in the Poetry file, does).

This isn't really the fault of UPM, since Python is the one not checking for these when it is trying to resolve the imports, however this needs to be worked around in order to be Universal until (if ever) Python fixes this, especially for versions that do not have the fixed version should it ever happen.

raxod502:
This should have been fixed by https://github.com/replit/upm/pull/6 cc @turbio but I guess it wasn't.

kajchang:
Still an issue for me.

jahands:
Looks like a user is having this same issue trying to use "pyjwt" instead of "jwt":

https://replit.com/talk/ask/Importing-issue-with-duplicate-import-names/132288

blast-hardcheese:
Looking into this, thanks for the report

PR #201, opened 2023-12-28
Ranked packages for python guess

blast-hardcheese:
Initial work towards #15 

- Converting `Guess` over to an import statement mapping to multiple packages that can provide that import
- If any package in the package array is installed, the entire array is removed
- After filtering, we pick the first package in every array and offer those as the response from `guess` 

Follow-on work to this as follows:
1) Augment the process that writes `internal/nix/python_map.json` to also include information about which package roots (the shallowest module in a package with more than one child, to start. There may be some utility in indexing what classes are available as well, but that'll come in the future)
2) Alter UPM to build a bespoke index by walking the `$PYTHONPATH` to exclude packages from `guess` that already supply necessary imports.

### DIFF
--- a/internal/backends/python/pypi_map.override.go
+++ b/internal/backends/python/pypi_map.override.go
@@ -18,5 +18,5 @@ var moduleToPypiPackageOverride = map[string][]string{
 	 */
 	"discord": {"discord.py"},
 	"bs4":     {"beautifulsoup4"},
-	"glm":     {"PyGLM"},
+	"glm":     {"PyGLM", "glm"},
 }
--- a/internal/backends/python/pypi_map.override.go
+++ b/internal/backends/python/pypi_map.override.go
@@ -3,20 +3,20 @@ package python
 /*
 Manual module -> package mapping overrides
 */
-var moduleToPypiPackageOverride = map[string]string{
-	"grpc_status":  "grpcio-status",       // 2nd most popular
-	"nvd3":         "python-nvd3",         // not popular enough 6th in popularity
-	"requirements": "requirements-parser", // popular rlbot depends on it, but doesn't supply requires_dist
-	"base62":       "pybase62",            // it was overridden by base-62 which wins due to name match but is less popular by far
-	"faiss":        "faiss-cpu",           // faiss is offered as precompiled wheels, faiss-cpu and faiss-gpu.
-	"graphics":     "graphics.py",         // this package is popular, but the module doesn't match the package name https://anh.cs.luc.edu/python/hands-on/3.1/handsonHtml/graphics.html#a-graphics-introduction
-	"replit.ai":    "replit-ai",           // Replit's AI package
+var moduleToPypiPackageOverride = map[string][]string{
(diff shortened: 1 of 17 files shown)


internal/backends/python/pypi_map.override.go
```diff
@@ -3,20 +3,24 @@ package python
...
-	"glm":     "PyGLM",
+var moduleToPypiPackageOverride = map[string][]string{
+	"grpc_status":  {"grpcio-status"},       // 2nd most popular
+	"nvd3":         {"python-nvd3"},         // not popular enough 6th in popularity
+	"requirements": {"requirements-parser"}, // popular rlbot depends on it, but doesn't supply requires_dist
+	"base62":       {"pybase62"},            // it was overridden by base-62 which wins due to name match but is less popular by far
+	"faiss":        {"faiss-cpu"},           // faiss is offered as precompiled wheels, faiss-cpu and faiss-gpu.
+	"graphics":     {"graphics.py"},         // this package is popular, but the module doesn't match the package name https://anh.cs.luc.edu/python/hands-on/3.1/handsonHtml/graphics.html#a-graphics-introduction
+	"replit.ai":    {"replit-ai"},           // Replit's AI package
+	"glm":          {"PyGLM", "glm"},        // Both of these packages are valid, but PyGLM is a library, glm is an executable.
```
cdmistman, line 14: i'm confused why this mapping is to a slice of strings, should we keep it 1:1 and keep it in the aliases mapping? i see `glm` is an executable, if you have it installed and `import glm` will that result in an error?

cdmistman: i don't think it makes sense to include it as an override, overrides should probably be a flat mapping and `glm` specifically is probably an alias instead

blast-hardcheese: Not always, consider if the developer of `glm` were to attempt to use Replit to develop the library, upm would suggest installing PyGLM.

Additionally, this opens us up to a world where we can have ranked or confidence-influenced guessing, based on classes or functions imported, not just absolute package names.

cdmistman: > Not always, consider if the developer of glm were to attempt to use Replit to develop the library, upm would suggest installing PyGLM.

isn't that what you've just implemented? i don't think that's a good idea - the developer is probably trying to import from their local namespace right? (if we suggest importing `glm` in this case we should absolutely address that 😬)

internal/backends/python/pypi_map.override.go
```diff
@@ -3,20 +3,24 @@ package python
...
+}
+
+/* Proxy packages
+ *
+ * These are packages that provide helpful aliases, but otherwise provide no functionality.
+ * We should prefer the real version.
+ */
+var moduleToPypiPackageAliases = map[string][]string{
```
cdmistman, line 22: pls make this `map[string]string`

cdmistman (approved):
after ^ addressed lgtm

PR #202, opened 2023-12-28
Recurse the current directory to find imports to guess

blast-hardcheese:
Somewhat surprising that we don't recurse already, but easy enough to implement.

### DIFF
--- a/internal/backends/nodejs/nodejs.go
+++ b/internal/backends/nodejs/nodejs.go
@@ -372,6 +372,10 @@ var nodejsGuessRegexps = util.Regexps([]string{
 	`(?m)(?:require|import)\s*\(\s*['"]([^'"{}]+)['"]\s*\)`,
 })
 
+var nodeIgnorePathSegments = map[string]bool{
+	"node_modules": true,
+}
+
 var jsPathGlobs = []string{
 	"*.js",
 	"*.jsx",
--- a/internal/util/tree-sitter.go
+++ b/internal/util/tree-sitter.go
@@ -43,6 +43,7 @@ func GuessWithTreeSitter(ctx context.Context, root string, lang *sitter.Language
 	var visited int
 	pathsToSearch := []string{}
 	err := fs.WalkDir(dirFS, ".", func(dir string, d fs.DirEntry, err error) error {
+		dir = path.Join(root, dir)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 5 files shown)


internal/util/tree-sitter.go
```diff
@@ -48,17 +48,26 @@ func GuessWithTreeSitter(ctx context.Context, dir string, lang *sitter.Language,
 	}
 
 	pathsToSearch := []string{}
-	for _, pattern := range searchGlobPatterns {
-		globSearchPaths, err := fs.Glob(dirFS, pattern)
+	err := fs.WalkDir(dirFS, ".", func(dir string, d fs.DirEntry, err error) error {
```
lhchavez, line 51: smol caveat: this will be slow af in some situations, where there are massive directories (like `node_modules` and `.pythonlibs` and `venv`) with gajillion files that would potentially match. we may be okay if we `return fs.SkipDir` if we encounter those directories to keep things speedy.

lhchavez: also can we have a smol test?

blast-hardcheese: So I was relying on the `ignored` infrastructure to resolve this, but you're right about if somebody does like a `mv node_modules node_modules.bak`. In doing some manual testing I discovered that the ignored stuff basically doesn't work anywhere near like what I expected, so I'm trying to figure out the best options we have. I'll add a test when I've got that sorted.

blast-hardcheese: OK! Ready for re-review

lhchavez (approved):

internal/backends/nodejs/nodejs.go
```diff
@@ -372,6 +372,10 @@ var nodejsGuessRegexps = util.Regexps([]string{
 	`(?m)(?:require|import)\s*\(\s*['"]([^'"{}]+)['"]\s*\)`,
 })
 
+var nodeIgnorePathSegments = map[string]bool{
+	"node_modules": true,
+}
+
```
masad-frost, line 378: Nice, I wonder if should load `.gitignore` and/or hardcode some of these folders https://github.com/github/gitignore/blob/main/Node.gitignore, and maybe ignore all directories that start with `.`?

masad-frost (approved):

internal/backends/nodejs/nodejs.go
blast-hardcheese: Yeah, I'd like to do that in a separate PR if you don't mind

PR #206, opened 2024-01-02
Overeager pip normalization

blast-hardcheese:
Unfortunately, even [normalization rules](https://packaging.python.org/en/latest/specifications/name-normalization/) are not the end of the story when it comes to `pip freeze`!

`pip install discord-py` will still result in `pip freeze | grep discord` == `discord.py`, since that's the name tracked in the metadata.

Exciting!

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -388,15 +388,24 @@ func makePythonPipBackend(python string) api.LanguageBackend {
 				util.Die("failed to run freeze: %s", err.Error())
 			}
 
+			// As we walk through the output of pip freeze,
+			// compare the package metadata name to the normalized
+			// pkgs that we are trying to install, to see which we
+			// want to track in `requirements.txt`.
+			normalizedPkgs := make(map[api.PkgName]bool)
+			for name := range pkgs {
+				normalizedPkgs[normalizePackageName(name)] = true
+			}
+
 			var toAppend []string
-			for _, line := range strings.Split(string(outputB), "\n") {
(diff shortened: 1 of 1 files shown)


internal/backends/python/python.go
```diff
@@ -368,6 +368,12 @@ func makePythonPipBackend(python string) api.LanguageBackend {
 			span, ctx := tracer.StartSpanFromContext(ctx, "pip install")
 			defer span.Finish()
 
+			normalizedPkgs := make(map[api.PkgName]api.PkgName)
+			for name := range pkgs {
+				// Deleting this denormalizes the output of upm list, fixing Workspace bug, but breaks upm add --lang pip
```
ryantm, line 373: > fixing Workspace bug

Can we reference or describe the bug so this isn't scary to change later?

ryantm (changes requested):

internal/backends/python/python.go
blast-hardcheese: I think that was a WIP comment, it doesn't read well. I'll fix it up, thanks

masad-frost:
python wtf moment

internal/backends/python/python.go
blast-hardcheese: Fleshed that out a bit. The underlying issue was Workspace was attempting to install `discord.py`, which we were normalizing and returning as `discord-py`, which would then (rightfully) confuse the workspace.

Hopefully the comment makes more sense.

ryantm (approved):

PR #209, opened 2024-01-03
bug/Differentiate exit codes

blast-hardcheese:
A decent first pass of trying to alert on UPM classifications seems to be to classify upm error categories. If one error code spikes, (`DieNetwork`, as an example,) we'll get a useful signal from that.

I didn't give much thought to the actual error codes, simply incrementing the previous code. If we want to do something clever like a HTTP status code `int(c / 100) == 2`-style pattern I'm happy to revisit, though that'll increase cardinality for minimal gain imo.


ryantm (approved):
Nice idea and looks good!

PR #211, opened 2024-01-04
chore/Avoid burning CI

blast-hardcheese:
We run `build`, then throw away the result and immediately rebuild inside `test`.

Removing the `build` phase will reduce our CI time by 4 minutes.


cdmistman (approved):
lmao the intention was to get the build artifact cached and it would be more clear what the result was but i couldn't get artifact caching to work 😢

blast-hardcheese:
When we get there, we can re-split -- I much prefer to build a binary and test it independently of the build environment. In the meantime, let's speed up CI 😁

PR #214, opened 2024-01-18
lock when there isnt a package dir either

cdmistman:
# Why

There's still some scenarios when `upm lock` doesn't happen. Right now the heuristic is:
1. `--force` flag is passed
2. lockfile doesn't exist
3. specfile has changed

We should also run if the package dir doesn't exist

# What Changed

if the 3 above conditions are all false, then also check to see if `backend.GetPackageDir()` returns a path, and if that path doesn't exist.

# Testing

test suite :)

for manual testing:
1. create a npm package
2. add a dependency and lock (however you choose to do this, doesn't matter)
3. `rm -rdf node_modules` if it exists
4. `upm lock` should recreate the `node_modules` dir

### DIFF
--- a/internal/cli/cmds.go
+++ b/internal/cli/cmds.go
@@ -216,7 +216,14 @@ func maybeLock(ctx context.Context, b api.LanguageBackend, forceLock bool) bool 
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


blast-hardcheese (approved):
makes sense

PR #216, opened 2024-01-19
chore/reflow gen_pypi_map cli

blast-hardcheese:
It was difficult to use this command because the CLI flags were all jumbled up together. I used the AI to split everything out into different functions, which was neat.

Tested by running various commands locally.


ryantm (approved):
:broom:

PR #217, opened 2024-01-19
chore/gen pypi map threshold

blast-hardcheese:
~Depends on #216~

Adding a minimum threshold for libraries to process.


ryantm (approved):

internal/backends/python/gen_pypi_map/package.go
```diff
@@ -0,0 +1,17 @@
...
+var normalizationPattern = regexp.MustCompile(`[-_.]+`)
+
+// normalizePackageName implements NormalizePackageName for the Python
+// backends.
+// See https://packaging.python.org/en/latest/specifications/name-normalization/
+func normalizePackageName(name string) string {
+	name = strings.ToLower(name)
+	name = normalizationPattern.ReplaceAllString(name, "-")
+	return name
+}
```
cdmistman, line 17: should probably share this code with the function that's already in `internal/backens/python/python.go`

cdmistman (approved):

internal/backends/python/gen_pypi_map/package.go
blast-hardcheese: It makes me think back to the OSS request of sharing our Go python tooling as a public lib. Perhaps someday, but this is still too low of a threshold to warrant that at this time, imo. I'm definitely thinking in that direction though!

PR #219, opened 2024-01-20
bug/python pip argspec

blast-hardcheese:
Why
===

When installing packages via pip by way of the `Packages` pane, we interpolate the CLI arg by way of `arg + spec`.

In looking for why this was the case, it turns out that we impose the same argument specifications on all languages, namely `<pkg> <spec>`, which gets confusing in situations like `flask[async]>=2,<3`, where the spec would actually be `[async]>=2,<3`.

image

Instead, we should defer to language specific parsing semantics where possible, accepting a wider variety of acceptable argspecs, which more closely resemble the args of the underlying commands, to be less confusing.

What changed
============

- Language-specific argument parsing overrides
- Sensible fallback of an empty version constraint to `==` instead of leaving it unconstrained
- For pip, maintain the users' version spec all the way through to writing back to `requirements.txt`, so `upm add flask>=2.3.0` doesn't get turned into `flask===2.3.3` on save.

Test plan
=========

```shell
upm -l python3 --ignored-packages unit_tests --ignored-paths .pythonlibs add 'flask 2'  # Fails, expectedly, since `==2` is not a valid version of flask
upm -l python3 --ignored-packages unit_tests --ignored-paths .pythonlibs add 'flask 2.3.3'
upm -l python3 --ignored-packages unit_tests --ignored-paths .pythonlibs add 'flask'
upm -l python3 --ignored-packages unit_tests --ignored-paths .pythonlibs add 'flask==2.3.3'
upm -l python3 --ignored-packages unit_tests --ignored-paths .pythonlibs add -- 'flask[async]>=2,<3'
```
these all work as expected in both Poetry and Pip

### DIFF
--- a/internal/api/types.go
+++ b/internal/api/types.go
@@ -19,6 +19,11 @@ func (n PkgName) HasPrefix(other PkgName) bool {
 	return strings.HasPrefix(string(n), string(other))
 }
 
+type PkgCoordinates struct {
+	Name string
+	Spec PkgSpec
+}
+
 // PkgSpec represents a package version constraint, e.g. "^1.1" or ">=
 // 1.1, <2.0" for Python.
 type PkgSpec string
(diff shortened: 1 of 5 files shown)


ryantm (approved):
Changes look good. I'm a little concerned you didn't have to touch any tests when making them.

blast-hardcheese:
> I'm a little concerned you didn't have to touch any tests when making them.

Turns out we do not have tests around invoking `upm add "$arg1" "$arg2" "$arg3"`. Good call.

blast-hardcheese:
image
:woohoo:

PR #224, opened 2024-01-23
bug/gen pypi map api incompat

blast-hardcheese:
Why
===

I didn't realize in #215 that BigQuery and Pypi shared the same data models, but had different APIs. Rectify that here.

What changed
============

- Duplicate pypi `PackageInfo` struct into `BqPackageInfo` for clarity
- Update README after #216 
- Give more useful countdown timer instead of just `5132152s 2days`

Test plan
=========


airportyh (approved):
Nice! Thanks.

PR #227, opened 2024-01-25
Found a typo; fixed it

airportyh:
Why
===

Tried loading all python sys deps into replit.nix, this one failed due to typo. Fixed it.


ryantm (approved):
Oops, thanks!

PR #231, opened 2024-02-08
Stripping out pkgs.glibc from imports

blast-hardcheese:
Why
===

[Reports online](https://discourse.nixos.org/t/conflict-between-glibc-and-glibclocales-tl-dr-i-messed-it-up/7608) suggest that glibc isn't intended to behave like a normal package, so we shouldn't automatically install it.

What changed
============

Avoid installing `pkgs.glibc`, as it was causing runtime conflicts

(example:
```
/nix/store/7q1b1bsmxi91zci6g8714rcljl620y7f-bash-5.2-p15/bin/bash: error while loading shared libraries: __vdso_gettimeofday: invalid mode for dlopen(): Invalid argument
```
)

Test plan
=========

Attempt to install any of the packages in this diff and ensure that sample code that uses that library runs as expected.


ryantm (approved):
Thanks, I think this makes sense!

PR #235, opened 2024-02-15
[bug] If packageDir does not exist, run Install

blast-hardcheese:
Why
===

During testing, `rm -rf .pythonlibs` followed by a `Run` skips package install completely:

Turns out this is because we don't verify the package directory actually exists
```shell
~/2024-02-15-venv-test$ upm show-package-dir
/home/runner/2024-02-15-venv-test/.pythonlibs
~/2024-02-15-venv-test$ du -sh $(upm show-package-dir)
du: cannot access '/home/runner/2024-02-15-venv-test/.pythonlibs': No such file or directory
```

What changed
============

- If the package directory isn't present, ignore the cache and run `install`.

Test plan
=========

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -429,7 +429,7 @@ func makePythonPipBackend(python string) api.LanguageBackend {
 
 			if outputB, err := util.GetCmdOutputFallible([]string{
 				"python",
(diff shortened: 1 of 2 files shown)


ryantm (approved):
Nice, I think I've hit this before long ago.

PR #237, opened 2024-02-26
Ensure that we handle the optionality of the trailing newline in requirements.txt

blast-hardcheese:
Why
===

The `requirements.txt` handling during `upm add` did no checking to see whether the file had a newline or not before adding one. This would lead to situations where attempting to `upm add dep2` to a file that already contained `dep1` would result in `dep1dep2`.

What changed
============

Determine whether the file already has a newline. If it does, maintain that newline as we add new lines to the file. If it doesn't, maintain its absence.

Test plan
=========

```
echo -n pytest > requirements.txt
upm add flask
```

Prior to this change, you'd end up with `pytestFlask\n`. After this change, we will end up with `pytest\nFlask`

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -495,13 +495,39 @@ func makePythonPipBackend(python string) api.LanguageBackend {
 				}
 			}
 
-			handle, err := os.OpenFile("requirements.txt", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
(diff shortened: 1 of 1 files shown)


ryantm (approved):
Thanks for the quick fix. Looks fine for Linux file endings. I expect this to be trouble for files with \r\n endings though. I'm not sure how many, if any, users we have on Windows.

blast-hardcheese:
We hardcode `\n` everywhere today, supporting Windows will require significant effort, this change doesn't preclude that in the future, so I'm gonna proceed.

PR #238, opened 2024-02-28
bug/Refreshing package map

blast-hardcheese:
Why
===

Main motivator here is that `upm guess` guesses `tensorflow-macos`

What changed
============

I figured that was good enough motivation to just refresh the package map.
Some packages dropped out due to being below the threshold for inclusion.

Test plan
=========

given the following in `main.py`:
```
import tensorflow as tf
print(tf.add(1, 2).numpy())
hello = tf.constant('Hello, TensorFlow!')
print(hello.numpy())
```

the following command...
```
...$ upm guess
tensorflow-macos
```
should not return the `-macos` variant

### DIFF
--- a/internal/backends/python/gen_pypi_map/gen_pypi_map.go
+++ b/internal/backends/python/gen_pypi_map/gen_pypi_map.go
@@ -83,11 +83,17 @@ func cmd_test(args []string) {
 		}
 		bqCache = normalizedBqCache
 
-		packageList := []string{}
+		// Deduplicate results
+		packageMap := make(map[string]bool)
 		for pkgName, count := range bqCache {
 			if count < *testThreshold {
 				continue
 			}
+			packageMap[pkgName] = true
+		}
+
+		packageList := []string{}
+		for pkgName := range packageMap {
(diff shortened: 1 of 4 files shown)


internal/backends/python/gen_pypi_map/gen_pypi_map.go
```diff
@@ -83,11 +85,40 @@ func cmd_test(args []string) {
...
 
-		packageList := []string{}
+		packageRemaps := make(map[string]string)
+		file, err := os.Open(*testRemapFile)
+
+		if err == nil {
+			scanner := bufio.NewScanner(file)
+			for scanner.Scan() {
+				columns := strings.SplitN(scanner.Text(), ",", 3) // old,new,description
+				if len(columns) > 0 {
```
ryantm, line 95:
```suggestion
                                old := columns[0]
                                new := columns[1]
				if len(columns) > 0 {
```
You might consider doing something like this to avoid the comment and keeping track of indexes but very minor.

ryantm (approved):

PR #239, opened 2024-03-01
bug/better tensorflow recommendations

blast-hardcheese:
Why
===

Follow-up to #238, tweak `guess` and search results to favor `tensorflow` instead of auxiliary packages. If users want those packages, they're more than welcome to install them directly, but upm should not suggest them.

What changed
============

- More aggressive filtering of search results
- Map `tensorflow-${arch}` and `tf-nightly-${arch}` to the common `tensorflow`

Test plan
=========

```
$ upm search Tensorflow  # tensorflow should be on the top
$ upm search Tensorflow1 # tensorflow should be on the top

$ echo 'import tensorflow' > main.py
$ upm guess # tensorflow
```


cdmistman (approved):
lgtm

PR #240, opened 2024-03-05
feat/python guess inside ifstat

blast-hardcheese:
Why
===

It's reasonably common for Python libraries to include gate imports behind `if`, `def`, and `try` statements. Let's try adding `if`/`elif`/`else` and see whether this is even a direction we want to proceed with.

What changed
============

Altered query to reflect the `if_statement` structure. Extended test to cover this new case.

Test plan
=========

See whether imports gated behind `if` statements are caught by `upm guess`

### DIFF
--- a/internal/backends/python/grab.go
+++ b/internal/backends/python/grab.go
@@ -19,7 +19,23 @@ var importsQuery = `
               name: (dotted_name) @import)])
 
    (import_from_statement
-     module_name: (dotted_name) @import)]
+     module_name: (dotted_name) @import)
+
+    (if_statement
+      condition: (_)
+      consequence: (block
+        [
+          (import_statement
+            name: [(dotted_name) @import
+               (aliased_import
+                 name: (dotted_name) @import)]
+          )
+          (import_from_statement
+            module_name: (dotted_name) @import)
+        ]
+      )
+    )
+  ]
 
   .
 
--- a/internal/backends/python/grab.go
+++ b/internal/backends/python/grab.go
@@ -13,54 +13,33 @@ import (
 
 var importsQuery = `
 (module
-  [
-    (import_statement
-      name: [
-        (dotted_name) @import
-        (aliased_import
-          name: (dotted_name) @import
-        )
-      ]
-    )
-
-    (import_from_statement
-      module_name: (dotted_name) @import
-    )
-
-    (if_statement [
-      (block [
-        (import_statement
-          name: [
-            (dotted_name) @import
+  [(import_statement
+     name: [(dotted_name) @import
             (aliased_import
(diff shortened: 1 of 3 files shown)


blast-hardcheese:
Before this gets merged, we should probably ensure that upm acknowledges poetry dependency groups. I don't know if it's worth attempting to guess alternative names for `requirements.txt`, though it may be beneficial to add support for `Pipfile`/`pipenv`.

I don't think `pip-tools` offers a huge lift, since `upm` already does enough of what `pip-compile` does today.

cdmistman:
> I don't know if it's worth attempting to guess alternative names for requirements.txt, though it may be beneficial to add support for Pipfile/pipenv.

i think as long as we support other dependency groups with poetry it should be fine. i think the majority of repls use poetry

internal/backends/python/grab.go
```diff
@@ -13,17 +13,54 @@ import (
...
-     name: [(dotted_name) @import
+  [
+    (import_statement
+      name: [
+        (dotted_name) @import
+        (aliased_import
+          name: (dotted_name) @import
+        )
+      ]
+    )
```
cdmistman, line 24: nit: please keep standard lisp conventions on ending parens

internal/backends/python/grab.go
```diff
@@ -13,17 +13,54 @@ import (
...
-              name: (dotted_name) @import)])
-
-   (import_from_statement
-     module_name: (dotted_name) @import)]
+              name: (dotted_name) @import
+            )
+          ]
+        )
+        (import_from_statement
+        module_name: (dotted_name) @import)
```
cdmistman, line 41:
```suggestion
          module_name: (dotted_name) @import)
```

cdmistman (commented):
please reformat with standard lisp formatting - this formatting is usually easier for the author to read but harder for others to read 😅

replit.nix
```diff
@@ -14,8 +14,8 @@
     pkgs.nodejs-18_x
     pkgs.yarn
     pkgs.nodePackages.pnpm
-    pkgs.python310Full
-    pkgs.python310Packages.pip
+    pkgs.python311Full
+    pkgs.python311Packages.pip
```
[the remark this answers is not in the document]
blast-hardcheese: Since before 2.7 afaik

blast-hardcheese:
> please reformat with standard lisp formatting

Open to any suggestion as to how to make maintaining query formatting easier. Are there standard formatters you recommend for folk that don't know what "standard lisp formatting" looks like?

cdmistman (approved):
lgtm!

airportyh (approved):

PR #241, opened 2024-03-07
feat/Filter dep groups out of guess

blast-hardcheese:
Why
===

As we prepare to merge #240, it will be good to filter out known-but-optional dependencies to avoid noise.

What changed
============

- Described poetry dependency groups
- Added a `mergeAllGroups bool` to `ListSpecfile` to include all dependency specs that we could find

Test plan
=========

```bash
$ poetry init -n
$ echo 'import flask' > main.py
$ upm guess  # Outputs "flask"
$ poetry add -G foo flask
$ upm guess  # No longer outputs anything
```

### DIFF
--- a/internal/cli/cmds.go
+++ b/internal/cli/cmds.go
@@ -312,7 +312,7 @@ func runAdd(
 
 	if util.Exists(b.Specfile) {
 		s := silenceSubroutines()
-		for name := range b.ListSpecfile() {
+		for name := range b.ListSpecfile(true) {
 			delete(normPkgs, b.NormalizePackageName(name))
 		}
 		s.restore()
@@ -359,7 +359,7 @@ func runRemove(language string, args []string, upgrade bool,
 	}
 
 	s := silenceSubroutines()
-	specfilePkgs := b.ListSpecfile()
+	specfilePkgs := b.ListSpecfile(false)
 	s.restore()
 
 	// Map whose keys are normalized package names.
@@ -464,7 +464,7 @@ func runList(language string, all bool, outputFormat outputFormat) {
 		var results map[api.PkgName]api.PkgSpec = nil
 		fileExists := util.Exists(b.Specfile)
 		if fileExists {
-			results = b.ListSpecfile()
+			results = b.ListSpecfile(false)
 		}
 		switch outputFormat {
 		case outputFormatTable:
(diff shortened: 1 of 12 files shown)


internal/cli/cmds.go
```diff
@@ -359,7 +359,7 @@ func runRemove(language string, args []string, upgrade bool,
 	}
 
 	s := silenceSubroutines()
-	specfilePkgs := b.ListSpecfile()
+	specfilePkgs := b.ListSpecfile(false)
```
cdmistman, line 362: why do we need to not merge all groups here?

blast-hardcheese: My presumption was that `remove` wouldn't remove from groups that weren't explicitly exposed, but that seems to not be the case. I'll flip this over!

blast-hardcheese:
```suggestion
	specfilePkgs := b.ListSpecfile(true)
```

blast-hardcheese: I tested `npm install --save-dev $pkg && npm remove $pkg` as well as `poetry add -G foo $pkg && poetry remove $pkg`, both passed. If there are edge cases we can pick them up as they come.

internal/cli/cmds.go
```diff
@@ -464,7 +464,7 @@ func runList(language string, all bool, outputFormat outputFormat) {
 		var results map[api.PkgName]api.PkgSpec = nil
 		fileExists := util.Exists(b.Specfile)
 		if fileExists {
-			results = b.ListSpecfile()
+			results = b.ListSpecfile(false)
```
cdmistman, line 467: maybe this should be true if the `-a` flag is specified?

blast-hardcheese: This is getting into the realm of "proper support for dependency groups", which isn't in scope for this change. I'm happy to make this `true`, though I don't feel great about a small scoped internal change leaking this much into the user's surface area.

cdmistman: happy to stamp but still think it makes sense when you're asking for _all_ dependencies

blast-hardcheese: Reasonable point.

blast-hardcheese:
```suggestion
			results = b.ListSpecfile(true)
```

PR #247, opened 2024-03-21
update go cache vendorHash

cdmistman:
Why
===

main immediately failed after merging #155

What changed
============

updated the vendorhash

Test plan
=========

ci passes

### DIFF
--- a/nix/upm/default.nix
+++ b/nix/upm/default.nix
@@ -7,7 +7,7 @@
   runCommand,
 }:
 let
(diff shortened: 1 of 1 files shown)


ryantm (approved):

blast-hardcheese:
Does vendorhash need to be updated every PR now, similar to locking in nixmodules?

cdmistman:
> Does vendorhash need to be updated every PR now, similar to locking in nixmodules?

no! just when you add/remove a dependency :)

PR #261, opened 2024-04-17
Skip upm lock when specfile lists no pkgs (DX-575)

alexhulbert:
Why
===

https://linear.app/replit/issue/DX-575/upm-lock-can-run-multiple-times-if-packagedir-doesnt-get-created

What changed
============

In the previous scenarios with no package folder (like .pythonlibs) where upm lock was always running, now upm will only run lock if the specfile exists and contains at least one package.

Test plan
=========

Tested out python template locally with no pythonlibs folder. I don't see packager messages when I don't change dependencies.


blast-hardcheese (approved):
Looks sensible!

ryantm (approved):

PR #267, opened 2024-05-29
Added entry for opencv-python dep on libgl

airportyh:
Why
===

opencv-python doesn't work on replit due to missing dep on libGL. https://replit.slack.com/archives/C03KS2B221W/p1716992546783999

There's a secondary issue: even with this fix, it still won't work if opencv needs to render UI. But that will be dealt with separately.

What changed
============

Added entry to in sysdep assistence for opencv-python->libGL.

Test plan
=========

1. create python repl
2. write `import cv2` in main.py
3. hit run. It should work


ink404 (approved):
Looks good, thank you!

PR #269, opened 2024-06-14
Adding system deps for manimgl

blast-hardcheese:
Why
===

Users have run into issues installing manimgl

What changed
============

Adding deps for manimgl

Test plan
=========


blast-hardcheese:
Looks like this never got merged, so here we are.

ryantm (approved):
Thank you!

PR #270, opened 2024-06-17
feat/flit support

blast-hardcheese:
Why
===

As we continue to expand into the realm of offering better Python package manager support, we're happy to add `flit` to upm.

What changed
============

- Bump `BurntSushi/toml` to 1.4.0
- Broke out `python/poetry.go` and `python/pip.go` to go along with `python/flit.go`
- Fix a bug with pip `requirements.txt` additions where we would take a zero-byte `requirements.txt` and turn it into `\ndep` instead of `dep\n`
- Adding `QuirksCannotAddRemove` to support `flit`, which does not directly support `add` or `remove`. In the future, we can or delete from `pyproject.toml`

Test plan
=========

- Import `https://github.com/mwaskom/seaborn` into Replit
- Run `pip install flit` to get a compatible version of flit
- Run `upm install`
- Verify that dependencies are installed


internal/backends/python/python.go
cdmistman, line 1: 🐐

internal/cli/cmds.go
```diff
@@ -267,6 +267,11 @@ func runAdd(
 	defer span.Finish()
 	b := backends.GetBackend(ctx, language)
 
+	if !b.QuirksCanAddRemove() {
+		fmt.Println("This backend does not suppport dependency additions")
```
cdmistman, line 271:
```suggestion
		fmt.Println("This backend does not suppport dependency addition")
```
nit to keep consistent with the other error message

cdmistman (changes requested):
integration tests please! lgtm so far though

blast-hardcheese:
A workaround was found for flit. This work can be brought back if necessary.

PR #272, opened 2024-06-25
Check error before using response

airportyh:
Why
===

We found a nil deref error in a template test that suggests this as the culprit.

https://replit.slack.com/archives/C06KAKG35QS/p1719352272057069?thread_ts=1719341894.359499&cid=C06KAKG35QS

What changed
============

Reversed the order of operations.


ryantm (approved):
thanks

PR #274, opened 2024-07-19
chore(flake): update build-go-cache for less rebuild

zimbatm:
Why
===

This is a small developer UX issue. In some cases, Nix would rebuild instead of fetching the result from the binary cache.

What changed
============

https://github.com/numtide/build-go-cache/pull/9 

Thanks to katexochen for fixing this issue.

Test plan
=========

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,17 +8,17 @@
         ]
       },
       "locked": {
-        "lastModified": 1699381303,
-        "narHash": "sha256-qi1FdxpIfHrx6T3W2ck7WOoP2B486H3WSRPFMPJtofs=",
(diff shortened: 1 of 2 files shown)


katexochen:
@zimbatm there is actually a workaround in place to prevent the rebuilds in another way (the comment pointed me to the problem in the first place):

https://github.com/replit/upm/blob/91c21f5ca1ae1e664f9c288515e332517c385f6d/nix/upm/default.nix#L16-L20

This could be removed now. :)

blast-hardcheese:
Based on the conversation here it seems this is not necessary, closing this out for now but please feel free to follow up if I misunderstood. Thank you!

PR #275, opened 2024-07-25
DX-780: Avoid guessing common words/module names

blast-hardcheese:
Why
===

pypi permits packages with extraordinarily common names ([utils](https://pypi.org/project/utils/)) as well as misconfigured packages (`gardener-cicd-base`, in this case, pollutes `site-packages` with `util.py`, `model/`, and `ci/`, among others)

What changed
============

I crossreferenced top-level packages discovered from pypi packages with `/usr/share/dict/words` then hand-pruned down to what I believe is a sensible list of names that can be common internal module names for projects, as well as excluding names that are popular pypi packages.

If there are any incorrect or missing entries, we can tune this over time, this is just trying to prevent `import util` guessing random stuff.

- Add a wordlist
- Delete guesses based on inclusion in the wordlist

Test plan
=========

### Before
```
$ echo $'import util\nimport utils' > main.py
$ upm guess --force
gardener-cicd-base
utils
```

### After
```
$ echo $'import util\nimport utils' > main.py
$ upm guess --force
$
```


cdmistman (approved):
nice

Issue #244, opened 2024-03-17
Add support for rye/ uv 


AYA-123:
Astral.sh makes an incredibly fast python package installer / resolver, that would be potentially very beneficial: https://github.com/astral-sh/uv , with 
> [10-100x faster](https://github.com/astral-sh/uv/blob/main/BENCHMARKS.md) than pip and pip-tools (pip-compile and pip-sync).

Potentially following on from this, ruff as a formatter would also be potentially beneficial - https://github.com/astral-sh/ruff -  being orders of magnitude times faster than existing linters ;/ formatters  like flake8 / black

PR #283, opened 2024-08-30
chore/Reflow test populator

blast-hardcheese:
Why
===

To support #244, do some housekeeping.

What changed
============

- Reorganize `python/python.go`. Reduce spaghetti by a bit.
- We shouldn't care about how to populate a project during tests. We know the files are embedded, we can recurse the directories and copy the files into a target. All this file path juggling is unnecessary.

Test plan
=========

CI should still pass


ryantm (approved):
Looks good, thanks!

PR #284, opened 2024-08-30
feat/uv support

### LINKED ISSUE #244: Add support for rye/ uv

blast-hardcheese:
Why
===

Resolve #244 

What changed
============

- Adding `python3-uv` backend

Test plan
=========

CI, manual testing


ryantm (approved):
Yay uv. We'll only use it for hatchling.build for now, so this seems safe and looks good to me!

blast-hardcheese:
I relaxed the requirement for `hatchling.build` as uv apparently only uses it as the default, there are CLI flags to override it. I altered it to "so long as Poetry is not configured", since we only support the three currently.

blast-hardcheese:
We're gradually rolling `uv` out to new Repls, starting with the agent. If you'd like to use it today, the easiest way is in a new repl:

1. Make sure you have a python nixmodule installed (Either fork the Python template, run `python` at the shell and type `Y` to add the module, or use the `Configuration` pane to select a version of Python to use).
2. `upm add -l python3-uv add flask` will create a new uv project, including making `.pythonlibs` a full virtualenv, and install flask. Subsequent operations inside Replit will correctly detect uv and use it.
3. Occasionally, after rebuilding your Repl environment (for instance when changing nix channels), you'll notice that uv will delete and recreate your `.pythonlibs` directory.
  This is because the nix store path for the `python` executable may have been removed in the upgrade.
  
  You can run `replit-fsck` after an upgrade to check whether any links or shebangs need to be patched, which should maintain your `.pythonlibs` even through an upgrade. This'll eventually be part of the full flow before we take uv GA.

Thanks for the interest!

PR #290, opened 2024-09-11
[dependabot] disable for python and node

ryantm:
Why
===
* We don't use python and node for upm, we use Go
* We're getting reports we don't need to address on dependencies in our test suite.
* It doesn't matter if there is an outdated or insecure dep in our test suite, since we aren't even running python or node code

What changed
===
* Limit dependabot to 0 PRs in pip and npm ecosystems

Test plan
===
* Stop seeing dependabot PRs for node and python

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,11 +6,13 @@ updates:
     package-ecosystem: "npm"
     schedule:
       interval: "monthly"
+    open-pull-requests-limit: 0
     ignore:
     - dependency-name: "*"
   - directory: "/test-suite"
     package-ecosystem: "pip"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,11 +6,13 @@ updates:
     package-ecosystem: "npm"
     schedule:
       interval: "monthly"
+    open-pull-requests-limit: 0
```
blast-hardcheese, line 9: why this instead of just deleting the whole `package-ecosystem` block for the languages we don't care about?

blast-hardcheese (approved):
Either way is fine, thanks for fixing this!

PR #294, opened 2024-09-27
bug/Proactively remove uv detritus

blast-hardcheese:
Why
===

`uv init` creates `README.md`, `hello.py`, and `.python-version` files. These are noisy and don't match the behavior of other python package managers in upm, so let's patch that up.

What changed
============

- Added more flags to `uv init`
- Test for `hello.py` before and after `uv init`. If it gets created through that process, delete it.

Test plan
=========

New agent runs should not have `hello.py`, uv's `README.md`, or `.python-version` files

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -4,6 +4,7 @@ package python
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -776,6 +777,13 @@ func makePythonUvBackend() api.LanguageBackend {
 			defer span.Finish()
 			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
+				// uv (currently?) creates a "hello.py" on uv init. Ensure it gets deleted before control returns to the user.
+				sampleFileName := "hello.py"
+				if util.Exists(sampleFileName) {
+					// Turns out it didn't exist after all!
+					sampleFileName = ""
+				}
+
 				cmd := []string{"uv", "init", "--no-progress", "--no-readme", "--no-pin-python"}
 
 				if projectName != "" {
@@ -783,6 +791,9 @@ func makePythonUvBackend() api.LanguageBackend {
 				}
 
 				util.RunCmd(cmd)
+				if sampleFileName != "" && util.Exists(sampleFileName) {
+					os.Remove(sampleFileName)
+				}
 			}
 
 			cmd := []string{"uv", "add"}
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -779,6 +779,7 @@ func makePythonUvBackend() api.LanguageBackend {
 			if !util.Exists("pyproject.toml") {
 				// uv (currently?) creates a "hello.py" on uv init. Ensure it gets deleted before control returns to the user.
 				sampleFileName := "hello.py"
+				// If the user already _has_ a file called hello.py, do not delete it for them.
 				if util.Exists(sampleFileName) {
 					// Turns out it didn't exist after all!
 					sampleFileName = ""
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -776,7 +776,7 @@ func makePythonUvBackend() api.LanguageBackend {
 			defer span.Finish()
 			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
-				cmd := []string{"uv", "init", "--no-progress"}
+				cmd := []string{"uv", "init", "--no-progress", "--no-readme", "--no-pin-python"}
 
 				if projectName != "" {
 					cmd = append(cmd, "--name", projectName)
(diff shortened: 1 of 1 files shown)


internal/backends/python/python.go
```diff
@@ -776,13 +777,25 @@ func makePythonUvBackend() api.LanguageBackend {
...
 
 				if projectName != "" {
 					cmd = append(cmd, "--name", projectName)
 				}
 
 				util.RunCmd(cmd)
+				if sampleFileName != "" {
+					if _, err := os.Stat(sampleFileName); errors.Is(err, os.ErrNotExist) {
+						os.Remove(sampleFileName)
+					}
```
cbrewster, line 797: Seems like the logic is inverted, right now we only delete the file if the file doesn't exit?

lhchavez, line 797: we can probably yolo-`os.Remove` the file.

lhchavez (approved):

internal/backends/python/python.go
```diff
@@ -776,13 +777,25 @@ func makePythonUvBackend() api.LanguageBackend {
 			defer span.Finish()
 			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
-				cmd := []string{"uv", "init", "--no-progress"}
+				// uv (currently?) creates a "hello.py" on uv init. Ensure it gets deleted before control returns to the user.
+				sampleFileName := "hello.py"
+				if _, err := os.Stat(sampleFileName); errors.Is(err, os.ErrNotExist) {
+					// Turns out it didn't exist after all!
+					sampleFileName = ""
+				}
```
ryantm, line 785: This looks like it is going to remove a user's hello.py file, because if it can stat the file, it won't be an error, and then it won't set sampleFileName to "" and skip it.

internal/backends/python/python.go
```diff
@@ -776,13 +777,25 @@ func makePythonUvBackend() api.LanguageBackend {
 			defer span.Finish()
 			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
-				cmd := []string{"uv", "init", "--no-progress"}
+				// uv (currently?) creates a "hello.py" on uv init. Ensure it gets deleted before control returns to the user.
+				sampleFileName := "hello.py"
+				if _, err := os.Stat(sampleFileName); errors.Is(err, os.ErrNotExist) {
```
lhchavez, line 782: oh wait, we also need to check whether `err` is nil. maybe

```suggestion
				if !util.Exists("hello.py") {
```

blast-hardcheese: `errors.Is(nil, os.ErrNotExist)` should be false, right? `nil` is not `ErrNotExist`

internal/backends/python/python.go
blast-hardcheese: Ah, right.

cbrewster, line 785: Isn't the file created by `uv init` itself? Right now it looks like if there is a `hello.py` (maybe from the user) before `uv init` is called, we'll go ahead and delete it after calling `uv init`:

1. `sampleFileName := "hello.py"`
2. The file exists, so we leave `sampleFileName` as `hello.py`
3. We run `uv init`
4. `sampleFileName` is not empty, so we remove the file (well we would except for the other issue I mentioned)

cbrewster:
Oh goodness review race condition :rofl:

internal/backends/python/python.go
blast-hardcheese: Yep, addressed

internal/backends/python/python.go
```diff
@@ -776,13 +777,23 @@ func makePythonUvBackend() api.LanguageBackend {
 			defer span.Finish()
 			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
-				cmd := []string{"uv", "init", "--no-progress"}
+				// uv (currently?) creates a "hello.py" on uv init. Ensure it gets deleted before control returns to the user.
+				sampleFileName := "hello.py"
+				if util.Exists(sampleFileName) {
```
blast-hardcheese, line 782:
```suggestion
				// If the user already _has_ a file called hello.py, do not delete it for them.
				if util.Exists(sampleFileName) {
```

internal/backends/python/python.go
```diff
@@ -776,13 +777,24 @@ func makePythonUvBackend() api.LanguageBackend {
 			defer span.Finish()
 			// Initalize the specfile if it doesnt exist
 			if !util.Exists("pyproject.toml") {
-				cmd := []string{"uv", "init", "--no-progress"}
+				// uv (currently?) creates a "hello.py" on uv init. Ensure it gets deleted before control returns to the user.
+				sampleFileName := "hello.py"
+				// If the user already _has_ a file called hello.py, do not delete it for them.
+				if util.Exists(sampleFileName) {
+					// Turns out it didn't exist after all!
```
blast-hardcheese, line 784:
```suggestion
```

internal/backends/python/python.go
```diff
@@ -4,6 +4,7 @@ package python
 import (
 	"context"
 	"encoding/json"
+	"errors"
```
blast-hardcheese, line 7:
```suggestion
```

PR #295, opened 2024-10-07
chore/add replit object storage to guess

blast-hardcheese:
Why
===

Attempting to guess this package failed, completing `replit`. Not great, let's fix that.

What changed
============

- Added `replit.object-storage`
- Updated README

Test plan
=========

Can we guess `import replit.object_storage` correctly?


ryantm (approved):
Thanks for improving the dev docs.

PR #312, opened 2024-12-06
Adding a sqlite index fallback

blast-hardcheese:
Why
===

Pypi is making some changes to their search index. If we have any issues, we fall back to best effort served from the sqlite database.

What changed
============

Adding a fallback search result that just serves the package names. TODO is to add download count, but that requires some more thought.

Test plan
=========

does `upm search -l python flask` work inside a repl?

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -248,6 +248,9 @@ func searchPypi(query string) []api.PkgInfo {
 		query = renamed[0]
 	}
 	results, err := SearchPypi(query)
+	if len(results) == 0 {
+		results, err = SearchLocalSqlite(query)
+	}
 	if err != nil {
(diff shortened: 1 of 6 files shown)


internal/backends/python/python.go
```diff
@@ -248,6 +248,9 @@ func searchPypi(query string) []api.PkgInfo {
 		query = renamed[0]
 	}
 	results, err := SearchPypi(query)
+	if len(results) == 0 {
```
ertan, line 251: shouldn't this be `if err != nil`

ryantm (approved):

internal/backends/python/python.go
blast-hardcheese: unfortunately no, since SearchPypi returns an empty array here

PR #320, opened 2024-12-19
[uv] use torch cpu instead of cuda by default

rmehri01:
Why
===

- by default, the CUDA version of `torch` is installled which takes up a bunch of space, yet we don't support GPUs on replit
  - we also have similar overrides for `tensorflow` for this reason: https://github.com/replit/upm/blob/a3cbf8fc89afc339ab3d0f5da5579d278078a1d6/internal/backends/python/pypi_map.override.go#L3
- this leads to deployments failing because their image size is too large, let's try using the cpu version instead

What changed
============

- unfortunately we can't do the override in the same way as `tensorflow` since the cpu version of `torch` isn't available on pypi
- instead, add extra indexes that will be checked before pypi, described here: https://docs.astral.sh/uv/configuration/indexes/

Test plan
=========

- `make upm` on a linux machine
- make a python project that uses `uv`
- do `upm add torch` and see the following output, notably `torch==2.5.1+cpu`:

```
--> uv add torch --index https://download.pytorch.org/whl/cpu --index-strategy unsafe-best-match
Resolved 11 packages in 513ms
Installed 10 packages in 135ms
 + filelock==3.16.1
 + fsspec==2024.10.0
 + jinja2==3.1.4
 + markupsafe==3.0.2
 + mpmath==1.3.0
 + networkx==3.4.2
 + setuptools==75.6.0
 + sympy==1.13.1
 + torch==2.5.1+cpu
 + typing-extensions==4.12.2
```

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"runtime"
 	"strings"
 
 	"github.com/BurntSushi/toml"
@@ -20,6 +21,23 @@ import (
 )
 
 var normalizationPattern = regexp.MustCompile(`[-_.]+`)
+
+type extraIndex struct {
+	// url is the location of the index
+	url string
+	// os is the operating system to override the index for, leave empty
+	// to override on any operating system
+	os string
+}
+
+var torchCpu = extraIndex{
+	url: "https://download.pytorch.org/whl/cpu",
+	os:  "linux",
+}
+
+var extraIndexMap = map[string][]extraIndex{
+	"torch": {torchCpu},
+}
 
 // this generates a mapping of pypi packages <-> modules
 // moduleToPypiPackage pypiPackageToModules are provided
@@ -751,6 +769,22 @@ func makePythonUvBackend() api.LanguageBackend {
 
 		return pkgs
 	}
+	addExtraIndexes := func(pkgName string) {
+		extraIndexes, ok := extraIndexMap[pkgName]
+		if ok {
+			uvIndex := os.Getenv("UV_INDEX")
+
+			for _, index := range extraIndexes {
+				if strings.HasPrefix(runtime.GOOS, index.os) {
+					uvIndex = index.url + " " + uvIndex
+				}
+			}
+
+			os.Setenv("UV_INDEX", uvIndex)
+		}
+		os.Setenv("UV_INDEX_STRATEGY", "unsafe-best-match")
+	}
+
 	b := api.LanguageBackend{
 		Name:     "python3-uv",
 		Specfile: "pyproject.toml",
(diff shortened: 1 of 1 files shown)


internal/backends/python/python.go
```diff
@@ -825,6 +841,17 @@ func makePythonUvBackend() api.LanguageBackend {
 
 				cmd = append(cmd, pep440Join(name, spec))
 			}
+
+			for _, index := range extraIndexes {
+				if strings.HasPrefix(runtime.GOOS, index.os) {
+					cmd = append(cmd, "--index", index.url)
```
blast-hardcheese, line 847: Were there issues using the envvar approach? This seems to add an additional index to search for every package install, instead of scoping it down to just when trying to install pytorch

rmehri01: Not sure if I'm doing it wrong but I think doing `UV_INDEX=pytorch=https://download.pytorch.org/whl/cpu uv add torch` also does the same thing as `--index` where it adds the index to the `pyproject.toml`. I think `--extra-index-url` doesn't but it shows a warning to use `index` instead

blast-hardcheese: Ah! Looking more closely at the docs, I completely agree with you.

I do think it would be cleaner for the user if we were to use the envvar approach, as that won't show up in the packager pane (consider that `uv add pytest` would end up with `--index .../pytorch/...` in the command that we echo back in the console). No doubt we'd end up with support tickets asking why pytorch is involved at all.

---

The one additional observation I will make is that no matter which strategy we use to supply the extra indices, if that index is _down_ then uv fails to resolve at all.

We should conditionally supply these values based on whether the package is currently being installed or is in the existing dependencies.

We can exclude transitive dependencies for now and share with support that by promoting pytorch to be a direct dependency may resolve issues for certain libraries. If this is a big issue we can brainstorm other strategies for conditionally enabling that index, or man-in-the-middle uv failures and switch strategies based on whether an attempted invocation is successful.

rmehri01: Sounds good! Updated to use the env var (prepending to it if necessary) and made a mapping from package names to extra indexes to use so that we only override on installing those packages or we have those packages when doing `uv sync`.

Also removed `--index-strategy unsafe-best-match` for now but let me know if you think we should keep it.

internal/backends/python/python.go
```diff
@@ -824,7 +857,10 @@ func makePythonUvBackend() api.LanguageBackend {
 				}
 
 				cmd = append(cmd, pep440Join(name, spec))
+
+				addExtraIndexes(string(name))
```
blast-hardcheese, line 861: Unfortunately we need to do the full `listUvSpecfile()` dance in `Add` as well, since we could be `uv add`-ing into a project that has unsync'd dependencies already in the project file

internal/backends/python/python.go
blast-hardcheese: Definitely need to put the `UV_INDEX_STRATEGY=unsafe-best-match` in there, otherwise we aren't guaranteed to get a functional version of pytorch (at least given my testing)

internal/backends/python/python.go
rmehri01: Ah, good point!

blast-hardcheese (approved):
Thanks!

PR #323, opened 2024-12-31
[uv] use torch cpu integration instead of env var

rmehri01:
Why
===

- #320 doesn't seem to work on replit, I think we're running into https://github.com/astral-sh/uv/issues/9647
- Using the [pytorch integration
](https://docs.astral.sh/uv/guides/integration/pytorch/#using-a-pytorch-index) with the latest nightly uv seems to fix the issue

What changed
============

- Check to see if torch is in the added packages or already in the spec file and if so, idempotently add the torch cpu override to `pyproject.toml` using `toml-editor`

Test plan
=========

- Running locally as well as on replit should work now:

image

### DIFF
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -4,12 +4,13 @@ package python
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"regexp"
-	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/BurntSushi/toml"
@@ -22,22 +23,7 @@ import (
 
 var normalizationPattern = regexp.MustCompile(`[-_.]+`)
 
-type extraIndex struct {
-	// url is the location of the index
-	url string
-	// os is the operating system to override the index for, leave empty
-	// to override on any operating system
-	os string
-}
-
-var torchCpu = extraIndex{
-	url: "https://download.pytorch.org/whl/cpu",
-	os:  "linux",
-}
-
-var extraIndexMap = map[string][]extraIndex{
-	"torch": {torchCpu},
-}
+var torchOverrides = []string{"torch", "torchvision"}
 
 // this generates a mapping of pypi packages <-> modules
 // moduleToPypiPackage pypiPackageToModules are provided
@@ -73,6 +59,12 @@ type pyprojectTOMLGroup struct {
 	Dependencies map[string]interface{} `json:"dependencies"`
 }
 
+type pyprojectUVIndex struct {
+	Name     string `json:"name"`
+	Url      string `json:"url"`
+	Explicit bool   `json:"explicit"`
+}
+
 // pyprojectTOML represents the relevant parts of a pyproject.toml
 // file.
 type pyprojectTOML struct {
@@ -95,6 +87,7 @@ type pyprojectTOML struct {
 		} `toml:"poetry"`
 		Uv *struct {
 			Sources map[string]interface{} `toml:"sources"`
+			Index   []pyprojectUVIndex     `toml:"index"`
 		} `toml:"uv"`
 	} `toml:"tool"`
 }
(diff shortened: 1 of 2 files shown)


internal/backends/python/python.go
```diff
@@ -769,20 +761,39 @@ func makePythonUvBackend() api.LanguageBackend {
...
 
-			os.Setenv("UV_INDEX", uvIndex)
+		pyprojectFile, err := os.OpenFile("pyproject.toml", os.O_APPEND|os.O_WRONLY, 0600)
+		if err != nil {
+			return
+		}
+		defer pyprojectFile.Close()
+
+		torchCPUOverride := `
+[tool.uv.sources]
```
blast-hardcheese, line 781: so the reason this is particularly unpleasant is that TOML doesn't permit duplicate table keys. `[[tool.uv.index]]` might be fine, since it's an array, but it may need to be colocated. `[tool.uv.sources]` is absolutely not OK, and will error if the sources are already defined and we add a duplicate block

rmehri01: I see, maybe can we just deserialize it like in `readPyproject`, add the torch overrides, and then serialize it to `pyproject.toml` then?

blast-hardcheese: the problem _there_ is that we can't deserialize unless we know the full structure, and then even if we put it back we'll nuke comments and formatting

blast-hardcheese: I was thinking we could use something like https://github.com/ceejbot/tomato in order to make this happen, but I haven't actually packaged it into the available runtime

internal/util/toml-editor.go
```diff
@@ -0,0 +1,77 @@
...
+	"io"
+	"os/exec"
+)
+
+// TomlEditorOp is the format of the JSON sent to toml-editor
+type TomlEditorOp struct {
+	Op              string `json:"op"`
+	Path            string `json:"path,omitempty"`
+	TableHeaderPath string `json:"table_header_path,omitempty"`
+	DottedPath      string `json:"dotted_path,omitempty"`
```
blast-hardcheese, line 16: `DottedPath` is synonymous with `Path`, I am trying to kill it

internal/util/toml-editor.go
```diff
@@ -0,0 +1,77 @@
...
+	Results []interface{} `json:"results"`
+}
+
+func TomlEditorIsAvailable() bool {
+	_, err := exec.LookPath("toml-editor")
+	return err == nil
+}
+
+func ExecTomlEditor(ops []TomlEditorOp) (*TomlEditorResponse, error) {
+	cmd := exec.Command("toml-editor", "--path", "pyproject.toml")
```
blast-hardcheese, line 33: please bubble out `pyproject.toml` here

internal/util/toml-editor.go
```diff
@@ -0,0 +1,77 @@
...
+
+	encoder := json.NewEncoder(stdin)
+	err = encoder.Encode(ops)
+	if err != nil {
+		return nil, fmt.Errorf("toml-editor error: %s", err)
+	}
+	err = stdin.Close()
+	if err != nil {
+		return nil, fmt.Errorf("toml-editor error: %s", err)
+	}
```
blast-hardcheese, line 55: This seems like it would be better suited for after we've consumed everything from `stdout`, to avoid obscure buffer-full scenarios

internal/backends/python/python.go
```diff
@@ -769,20 +762,77 @@ func makePythonUvBackend() api.LanguageBackend {
...
+		}
+
+		// if the value isn't already present, add it
+		valueBytes, err := json.Marshal(value)
+		if err != nil {
+			return err
+		}
+		addOp := util.TomlEditorOp{
+			Op:              "add",
+			TableHeaderPath: fmt.Sprintf("%s/[[]]", path),
```
blast-hardcheese, line 802: `[[]]` is only applicable in the `tool/uv/index` case, not the `tool/uv/sources/%s` case

blast-hardcheese (commented):
Nearly there! You should also be able to test this in a repl, now that the latest `toml-editor` has rolled out 💪

internal/backends/python/python.go
rmehri01: Sorry why is that? From my understanding the `[[]]` is just syntactic sugar so we could also write `tool.uv.sources` and `tool.uv.index` the opposite way around?

```toml
[[tool.uv.sources.torch]]
index = "pytorch-cpu"
marker = "platform_system == 'Linux'"

[[tool.uv.sources.torchvision]]
index = "pytorch-cpu"
marker = "platform_system == 'Linux'"

[tool.uv]
index = [{ name = "pytorch-cpu", url = "https://download.pytorch.org/whl/cpu", explicit = true }]
```

blast-hardcheese: Hilariously, "arrays", "tables", and ["arrays of tables"](https://toml.io/en/v1.0.0#array-of-tables) are all distinct structures in toml. `[[]]` only applies to "Arrays of Tables", which is `tool.uv.sources`.

You can see in what you pasted above, you've got `[[tool.uv.sources.torch]]` which is invalid, as it should at the very least be `[...]`, or alternately `TableHeaderPath: tool/uv/sources` combined with `path = torch` which would get you the exact example from the docs:

```toml
[tool.uv.sources]
torch = [
  { index = "pytorch-cpu" },
]
torchvision = [
  { index = "pytorch-cpu" },
]
```

you've also got `tool.uv.index` backwards from what they list in the docs, as it's expected to be an array of tables, a la `[[tool.uv.index]]`. 

Depending on the parser, some or all of these concepts may be able to be swapped around, but given what `toml-editor` can do today it is possible to recreate the exact construct from the [uv documentation](https://docs.astral.sh/uv/guides/integration/pytorch/#using-a-pytorch-index)

internal/backends/python/python.go
```diff
@@ -769,20 +762,84 @@ func makePythonUvBackend() api.LanguageBackend {
...
+
+					url, ok := gotValue["url"]
+					if !ok {
+						return false
+					}
+
+					return url == "https://download.pytorch.org/whl/cpu"
+				}) {
+					return nil
 				}
```
blast-hardcheese, line 803: As much as it pains me to say this, the `ContainsFunc` is actually harder to read. I'd suggest:

```suggestion
				for value, _ := range arr {
					index, ok := value.(map[string]interface{})
					if !ok {
						continue
					}

					name, ok := index["name"]
					if !ok {
						continue
					}

					hasIndex := name == "pytorch-cpu"
					break
				}
```

internal/backends/python/python.go
```diff
@@ -769,20 +762,84 @@ func makePythonUvBackend() api.LanguageBackend {
...
+		}
+		addOp := util.TomlEditorOp{
+			Op:              "add",
+			TableHeaderPath: "tool/uv/index/[[]]",
+			Value:           string(valueBytes),
+		}
+		_, err = util.ExecTomlEditor("pyproject.toml", []util.TomlEditorOp{addOp})
+		if err != nil {
+			return err
+		}
```
blast-hardcheese, line 820: If we don't have the index, we should add it. Otherwise, we still need to check that we're actually pinning all the expected packages

```suggestion
		if !hasIndex {
			value := map[string]interface{}{"name": "pytorch-cpu", "url": "https://download.pytorch.org/whl/cpu", "explicit": true}
			valueBytes, err := json.Marshal(value)
			if err != nil {
				return err
			}
			addOp := util.TomlEditorOp{
				Op:              "add",
				TableHeaderPath: "tool/uv/index/[[]]",
				Value:           string(valueBytes),
			}
			_, err = util.ExecTomlEditor("pyproject.toml", []util.TomlEditorOp{addOp})
			if err != nil {
				return err
			}
		}
```

internal/backends/python/python.go
```diff
@@ -769,20 +762,84 @@ func makePythonUvBackend() api.LanguageBackend {
...
 
-			os.Setenv("UV_INDEX", uvIndex)
+		for _, name := range torchOverrides {
+			value := []map[string]interface{}{
+				{"index": "pytorch-cpu", "marker": "platform_system == 'Linux'"},
+			}
+			valueBytes, err := json.Marshal(value)
```
blast-hardcheese, line 826: We need to query the `pyproject.toml` `[tool.uv.sources]` table to see if each package in `torchOverrides` actually has a pinned value, or if we need to add it.

The reason being, if we need to add more pins in the future we should be able to operate with files that have never been modified, or ones that we've touched with a previous version of this approach.

rmehri01: Got it, did something similar to what you suggested above and tested adding a new torch override which seems to work!

blast-hardcheese (approved):
Works great!

PR #326, opened 2025-01-28
[python] Just add all the packages that transitively depend on torch for now

blast-hardcheese:
Why
===

Until uv's transitive-package-index-pinning is resolved, let's just proactively ensure we pin `torch` for anything that _will_ transitively depend on it.

This list was generated using:

```shell
cat internal/backends/python/pkgs.json | jq -c --slurp 'map(select(.requires_dist // [] | contains(["torch"]))) | map(.name)'
```

What changed
============

Exhaustively populate `torchOverrides.go`

Test plan
=========

Rough performance testing:

```
diff --git a/internal/backends/python/python.go b/internal/backends/python/python.go
index 8af0f13b..012b7e41 100644
--- a/internal/backends/python/python.go
+++ b/internal/backends/python/python.go
@@ -12,6 +12,7 @@ import (
 	"regexp"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/BurntSushi/toml"
 	"github.com/replit/upm/internal/api"
@@ -727,6 +728,28 @@ func makePythonPipBackend() api.LanguageBackend {
 
 // makePythonUvBackend returns a backend for invoking uv.
 func makePythonUvBackend() api.LanguageBackend {
+
+	before := time.Now()
+	i := 0
+	c := 0
+	seen := make(map[string]bool)
+	for i < 100000 {
+		pkgname := "zuko"
+		fst := pkgname[0:i % 4]
+		snd := pkgname[i % 4:len(pkgname)]
+		if _, ok := seen[fst]; !ok {
+			fmt.Println("Seen: ", fst, "+", snd)
+			seen[fst] = true
+		}
+		if slices.Contains(torchOverrides, string(fst + snd)) {
+			c += 1
+		}
+		i += 1
+	}
+	after := time.Now()
+
+	fmt.Println("100,000 runs:", after.Sub(before))
+
 	listUvSpecfile := func() map[api.PkgName]api.PkgSpec {
 		cfg, err := readPyproject()
 		if err != nil {
```

on a live repl in production yields acceptable performance results:
```
Seen:   + zuko
Seen:  z + uko
Seen:  zu + ko
Seen:  zuk + o
100,000 runs: 94.217053ms
```


ryantm (approved):
I :heart: big static lists.

PR #345, opened 2025-03-31
add support for installing .replit nix pkgs

vlinkz:
Why
===

Replit will support quickly installing nix packages using `nix.packages` in .replit, so lets make upm have an option to use that.

What changed
============

Use toml-editor to install python system dependency using `.replit` `nix.packages` field.

Test plan
=========

Test upm binary and check that it modifies `.replit` correctly given the installed python package.


internal/backends/python/python.go
```diff
@@ -1038,6 +1118,12 @@ func makePythonUvBackend() api.LanguageBackend {
 			specfilePkgs := listUvSpecfile()
 			commonInstallNixDeps(ctx, pkgs, specfilePkgs)
 		},
+		InstallDotReplitSystemDependencies: func(ctx context.Context, pkgs []api.PkgName) {
+			// Ignore the error here, because if we can't read the specfile,
+			// we still want to add the deps from above at least.
+			specfilePkgs := listUvSpecfile()
```
blast-hardcheese, line 1124: what error?

internal/backends/python/python.go
```diff
@@ -304,6 +304,74 @@ func commonInstallNixDeps(ctx context.Context, pkgs []api.PkgName, specfilePkgs
...
+
+	for _, pkg := range pkgs {
+		deps := nix.PythonNixDeps(string(pkg))
+		for _, dep := range deps.Deps {
+			dep = strings.TrimPrefix(dep, "pkgs.")
+			allPkgs[dep] = true
+		}
+	}
+
+	for pkg := range specfilePkgs {
```
blast-hardcheese, line 345: no nil check makes for a good time

internal/backends/python/python.go
```diff
@@ -304,6 +304,74 @@ func commonInstallNixDeps(ctx context.Context, pkgs []api.PkgName, specfilePkgs
...
+		deps := nix.PythonNixDeps(string(pkg))
+		for _, dep := range deps.Deps {
+			dep = strings.TrimPrefix(dep, "pkgs.")
+			allPkgs[dep] = true
+		}
+	}
```
blast-hardcheese, line 351: Hardcoding python here seems undesirable in a generic function

internal/backends/python/python.go
```diff
@@ -304,6 +304,74 @@ func commonInstallNixDeps(ctx context.Context, pkgs []api.PkgName, specfilePkgs
...
+		Op:   "get",
+		Path: "nix/packages",
+	}
+	response, err := util.ExecTomlEditor(".replit", []util.TomlEditorOp{getOp})
+	if err != nil {
+		util.DieSubprocess("toml-editor error: %s", err)
+	}
+	if len(response.Results) != 1 {
+		util.DieSubprocess("expected one result")
+	}
```
blast-hardcheese, line 324: Is zero not valid? Would this not explode on any repl without this field configured?

blast-hardcheese (approved):
One nit, one issue that needs to be fixed

internal/backends/python/python.go
vlinkz: `specfilePkgs` is an input of type `map[api.PkgName]api.PkgSpec`, so it's `nil` should just be an empty map and have 0 iterations right?

vlinkz: go does what go does (range does it's own nil check)

internal/backends/python/python.go
vlinkz: So apparently `toml-editor` will return `{"status":"success","message":null,"results":[null]}` if it does not exist, so there should still be 1 result

vlinkz: Also built and uploaded to a repl to confirm this

PR #357, opened 2025-06-13
[python] don't attempt systemdep install if no deps

vlinkz:
Why
===

_Describe what prompted you to make this change, link relevant resources: tasks, reports, discussions, etc

If no system dependencies are found, don't to toml-editor nop.

What changed
============

_Describe what changed to a level of detail that someone with no context with your PR could be able to review it_
Check if no system dependencies need to be installed for python packages.

Test plan
=========


vlinkz:
Hm this is not actually fixing it in the right place. I'll fix in a min, but we should check if no _new_ packages are added, not no packages at all.

